PR #6, opened 2022-01-12
Enable CI on Windows (and macOS)

kozross:
This extends our Github Actions support to both macOS (trivially) and Windows (_very_ far from trivially). Along the way, we also resolve the 'how the actual hell do we make this all work on Windows' issue, although it's not necessarily the most correct or ideal: see `Setup.hs` for details.


Kleidukos (approved):
Very good, thank you very much.

PR #11, opened 2022-01-18
Bindings to comparisons

kozross:
This provides bindings to constant-time comparison functions in libsodium. These are needed for basically everything to come later.


kozross:
So, because of [long string of curse words] Windows, we have to vendor all the headers too, along with`version.h`. We do have to take some care that it matches what the Autotools would spit out, but this is relatively minor.

PR #19, opened 2022-03-17
Add Generic and Passsword hashing

Kleidukos:
This PR adds bindings to the Generic and Password hashing APIs of Libsodium

Things to check for:

* [x] Documentation

### DIFF
--- a/src/Cryptography/Sodium/Bindings/Hashing.hs
+++ b/src/Cryptography/Sodium/Bindings/Hashing.hs
@@ -1,0 +1,215 @@
+{-# LANGUAGE CApiFFI #-}
+{-# LANGUAGE Trustworthy #-}
+
+-- |
+--
+-- Module: Cryptography.Sodium.Bindings.Hashing
+-- Description: Direct bindings to the generic hashing primitives of Libsodium.
+-- Copyright: (C) Hécate Moonlight 2022
+-- License: BSD-3-Clause
+-- Maintainer: The Haskell Cryptography Group
+-- Stability: Stable
+-- Portability: GHC only
+module Cryptography.Sodium.Bindings.Hashing
+  ( -- * Introduction
+    -- $introduction
+
+    -- * Operations
+    CryptoGenericHashState,
+    cryptoGenericHash,
+    cryptoGenericHashKeyGen,
+    withGenericHashState,
+    withGenericHashStateOfSize,
+    cryptoGenericHashInit,
+    cryptoGenericHashUpdate,
+    cryptoGenericHashFinal,
+
+    -- * Constants
+    cryptoGenericHashBytes,
+    cryptoGenericHashBytesMin,
+    cryptoGenericHashBytesMax,
+    cryptoGenericHashKeyBytes,
+    cryptoGenericHashKeyBytesMin,
+    cryptoGenericHashKeyBytesMax,
+  )
+where
+
+import Foreign (Ptr, allocaBytes)
+import Foreign.C (CInt (..), CSize (..), CUChar (..), CULLong (..))
+
+-- $introduction
+-- This API computes a fixed-length fingerprint for an arbitrarily long message.
+-- It is backed by the BLAKE2b algorithm.
+--
+-- Sample use cases:
+--
+--   * File integrity checking
+--   * Creating unique identifiers to index arbitrarily long data
+--
+-- ⚠ Do not use this API module to hash passwords!
+--
+-- Whenever there is a @'Ptr' 'CryptoGenericHashState'@, it must point to enough memory
+-- to hold the hash state.
+-- This is at least 'cryptoGenericHashBytesMin', at most
+-- 'cryptoGenericHashBytesMax', and should typically be 'cryptoGenericHashBytes'.
+-- It is the caller's responsibility to ensure that this holds.
+
+-- | Opaque tag representing the hash state struct @crypto_generichash_state@ used by the C API.
+--
+-- To obtain a 'CryptoGenericHashState', use 'withGenericHashState'.
+--
+-- @since 0.0.1.0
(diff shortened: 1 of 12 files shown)


src/Cryptography/Sodium/Bindings/Hashing.hs
```diff
@@ -0,0 +1,178 @@
...
+-- | This function allocates a 'CryptoGenericHashState' of the desired size.
+--
+-- Use the following constants as parameter to this function:
+--
+--   * 'cryptoGenericHashBytesMin' (16U)
+--   * 'cryptoGenericHashBytes' (32U)
+--   * 'cryptoGenericHashBytesMax' (64U)
+--
+-- @since 0.0.1.0
+allocateGenericHashStateOfSize :: CSize -> IO (Ptr CryptoGenericHashState)
```
kozross, line 78: I don't think this functionality should be provided here. You might want to use `allocaBytes` instead, or put it in a `ForeignPtr`.

Furthermore, the documentation is not entirely accurate: the min and max are _bounds_, so anything with size between them will also work.

src/Cryptography/Sodium/Bindings/Hashing.hs
```diff
@@ -0,0 +1,178 @@
...
+
+-- | Haskell binding to the @crypto_generichash_BYTES_MAX@ constant
+--
+-- It is a 64-bit unsigned integer (64U).
+--
+-- @since 0.0.1.0
+foreign import capi "sodium.h value crypto_generichash_BYTES_MAX"
+  cryptoGenericHashBytesMax :: CSize
+
+-- | Haskell binding to the @crypto_generichash_KEYBYTES@ constant
```
kozross, line 155: `CSize` is not necessarily 32 bits wide. According to the C standard, it is _at least_ 32 bits wide, but could be larger. This information is both redundant (we have the type right there) and misleading.

src/Cryptography/Sodium/Bindings/Hashing.hs
```diff
@@ -0,0 +1,178 @@
...
+-- with 'cryptoGenericHashFinal'.
+--
+-- @since 0.0.1.0
+foreign import capi "sodium.h crypto_generichash_final"
+  cryptoGenericHashFinal ::
+    -- | The hash state used throughout the previous hashing operations
+    Ptr CryptoGenericHashState ->
+    -- | The pointer to the resulting fingerprint
+    Ptr CUChar ->
+    CSize ->
```
kozross, line 128: What does this argument mean? We have to document every single one.

src/Cryptography/Sodium/Bindings/Hashing.hs
```diff
@@ -0,0 +1,178 @@
...
+    IO CInt
```
kozross, line 129: What value(s) can this return, and what do they mean?

kozross (changes requested):

src/Cryptography/Sodium/Bindings/PasswordHashing.hs
```diff
@@ -0,0 +1,275 @@
...
+    -- | @opslimit@ parameter. It represents the maximum amount of computations to perform. Raising this number will make the function require more CPU cycles
+    -- to compute a key. This number must be between 'cryptoPWHashOpsLimitMin' and 'cryptoPWHashOpsLimitMax'.
+    CULLong ->
+    -- | @memlimit@ parameter. It is the maximum amount of RAM in bytes that the function will use.
+    -- This number must be between 'cryptoPWHashMemLimitMin' 'cryptoPWHashMemLimitMax'
+    CSize ->
+    -- | @alg@ parameter. It is an identifier for the algorithm to use and should be set to one of the following values:
+    --     'cryptoPWHashAlgDefault', 'cryptoPWHashAlgArgon2I13' or 'cryptoPWHashAlgArgon2ID13'.
+    CInt ->
+    IO CInt
```
kozross, line 105: What does this return and what does it mean?

test/HashingTests.hs
```diff
@@ -0,0 +1,90 @@
...
+spec :: TestTree
+spec =
+  testGroup
+    "Hashing tests"
+    [ testCase "Test hashing a single message with a key" testHashingWithAKey,
+      testCase "Test hashing a single message with no key" testHashingWithNoKey
+    ]
+
+testHashingWithAKey :: Assertion
+testHashingWithAKey = do
```
kozross, line 23: In this setting, I would recommend the use of `allocaBytes` instead of `mallocBytes`. You don't need the memory to live on past this function call, and as-current, you leak memory.

test/HashingTests.hs
```diff
@@ -0,0 +1,90 @@
...
+    inLen
+    keyParameter
+    keyLen
+
+  result <- memcmp outParameter1 outParameter2 cryptoGenericHashBytes
+
+  assertBool "The result is expected" (result == 0)
+
+testHashingWithNoKey :: Assertion
+testHashingWithNoKey = do
```
kozross, line 63: Same here.

test/HashingTests.hs
```diff
@@ -0,0 +1,90 @@
...
+    outParameter
+    outLen
+    inParameter
+    inLen
+    keyParameter
+    keyLen
+
+  dataArray <- peekArray0 0x0 outParameter
+  let result = toEnum @Char . fromEnum <$> dataArray
+  assertEqual "The result is expected" result "X\181\149\168\&9\179\181i\191\146*\223\222U9\190\162\229l;\251\230\b\162k\162/\248!\170H\ACK"
```
kozross, line 87: Instead of round-tripping through `Char`, I think you can get a much more meaningful result by making an array to compare with.

kozross (changes requested):

PR #21, opened 2022-03-29
Fix CI (again)

kozross:
This should both optimize the CI (since we no longer have to build `ormolu`), and deal with the issue of failing hashes.


kozross:
This does not fix our Windows build issues, but that's a separate PR.

PR #22, opened 2022-03-29
ED25519 Signatures

Kleidukos:
This PR adds bindings to the ED25519 and SHA512 signature and hashing APIs of libsodium.

## TODO

* [x] Backlinks to C functions in the libsodium docs
  * [x] SHA256 
  * [x] SHA512
  * [x] ED25519

### DIFF
--- a/src/Cryptography/Sodium/Bindings/SHA2.hs
+++ b/src/Cryptography/Sodium/Bindings/SHA2.hs
@@ -87,7 +87,7 @@ foreign import capi "sodium.h crypto_hash_sha256"
 -- @since 0.0.1.0
 data CryptoHashSHA256State
 
--- | Perform an operation with a 'CryptoHashSHA256State of size 'cryptoHashSHA256StateBytes'
+-- | Perform an operation with a 'CryptoHashSHA256State' of size 'cryptoHashSHA256StateBytes'
 -- allocated and deallocated automatically.
 --
 -- ⚠ The return value of 'withCryptoHashSHA256State __MUST NOT__ leak the 'CryptoHashSHA256State'.
--- a/src/Cryptography/Sodium/Bindings/GenericHashing.hs
+++ b/src/Cryptography/Sodium/Bindings/GenericHashing.hs
@@ -86,7 +86,7 @@ withGenericHashStateOfSize size action = allocaBytes (fromIntegral size) action
 -- However, for specific use cases, the size can be any value between 'cryptoGenericHashBytesMin' (included)
 -- and 'cryptoGenericHashBytesMax' (included).
 --
--- The @key@ parameter can be NULL and keylen can be 0. In this case, a message will always have the same fingerprint
+-- The @key@ parameter can be 'Foreign.nullPtr' and keylen can be 0. In this case, a message will always have the same fingerprint
 -- But a key can also be specified. A message will always have the same fingerprint for a given key, but different
 -- keys used to hash the same message are very likely to produce distinct fingerprints.
 -- In particular, the key can be used to make sure that different applications generate different fingerprints even


src/Cryptography/Sodium/Bindings/SHA512.hs
```diff
@@ -0,0 +1,81 @@
...
+-- looking for a generic hash function and not specifically SHA-2, using
+-- 'Cryptography.Sodium.Bindings.GenericHashing' (BLAKE2b) might be a better choice. These functions
+-- are also not suitable for hashing passwords or deriving keys from passwords. Use
+-- 'Cryptography.Sodium.Bindings.PasswordHashing' instead.
+--
+-- These functions are not keyed and are thus deterministic. In addition, the untruncated versions
+-- are vulnerable to length extension attacks. A message can be hashed in a single pass, but a
+-- streaming API is also available to process a message as a sequence of multiple chunks.
+
+data CryptoHashSHA512State = CryptoHashSHA512State
```
kozross, line 40: Does this correspond to anything? If so, it should have a Haddock with a link, along with an explanation.

This applies to multiple other entries in this module.

src/Cryptography/Sodium/Bindings/SHA512.hs
```diff
@@ -0,0 +1,81 @@
...
+    count :: Word64,
+    buf :: Word8
+  }
+
+foreign import capi "sodium.h value crypto_hash_sha512_BYTES"
+  cryptoHashSHA512Bytes :: CSize
+
+foreign import capi "sodium.h crypto_hash_sha512"
+  cryptoHashSHA512 ::
+    -- | @out@ parameter. Cannot be @NULL@.
```
kozross, line 51: I assume these are the names of the C parameters? This isn't really very useful, for a few reasons:

* I have no idea what function you're documenting from reading the Haddocks, since nothing is linked; and
* Just naming the parameter tells me nothing about what it means. In many cases, this is _not_ obvious_: for example, what does it mean when `out` is `NULL`?
* `NULL` is a C concept: the Haskell equivalent is `nullPtr`.
* The function returns `IO CInt` - what does any given result mean?

This applies to multiple other functions.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,276 @@
...
+-- Combined Mode --
+-------------------
+
+-- | The 'cryptoSign' function prepends a signature to a message @m@, whose length is @mlen@ bytes,
+-- using the secret key @sk@.
+--
+-- The signed message, which includes the signature plus an unaltered copy of the message, is put
+-- into @sm@ and is 'cryptoSignBytes' + @mlen@ bytes long.
+--
+-- If @smlen@ is not a @NULL@ pointer, then the actual length of the signed message is stored in
```
kozross, line 104: I think 'If `smlen` is not `nullPtr`' makes far more sense in a Haskell context.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,276 @@
...
+-- @smlen@.
+--
+-- @since 0.0.1.0
+foreign import capi "sodium.h crypto_sign"
+  cryptoSign ::
+    -- | @sm@ parameter.
```
kozross, line 110: While clearer than the corresponding documentation in `Cryptography.Sodium.Bindings.SHA512`, I think this breaks a bit with convention in prior modules, where parameters are described 'inline' with a Haddock on the parameter, rather than being named, then described in the Haddock for the entire function.

If you think this style is better (and FWIW, I don't), we should be consistent, and pin this in place with a  note in our standards.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,276 @@
...
+    IO CInt
+
+-------------------
+-- Detached Mode --
+-------------------
+
+-- | The 'cryptoSignDetached' function signs the message @m@, whose length is @mlen@ bytes,
+-- using the secret key @sk@ and puts the signature into @sig@, which can be up to
+-- 'cryptoSignBytes' bytes long.
+-- The actual length of the signature is put into @siglen@ if @siglen@ is not @NULL@.
```
kozross, line 153: Again, best refer to `nullPtr` IMHO, as this is Haskell, not C.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,276 @@
...
+-------------------------
+-- Multi-part messages --
+-------------------------
+
+-- $mpm
+-- If the message you're trying to sign doesn't fit in memory, then it can be provided as a sequence of arbitrarily-sized
+-- chunks. This uses the @Ed25519ph@ signature system, which pre-hashes the message. In other words,
+-- what gets signed is not the message itself but its image through a hash function. If the message
+-- can fit in memory and be supplied as a single chunk, then the single-part API should be
+-- preferred. Note: @Ed25519ph(m)@ is intentionally not equivalent to Ed25519(SHA512(m)). If, for
```
kozross, line 182: This note should be a standalone section: `= Note`, more specifically.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,276 @@
...
+-- some reason, you need to pre-hash the message yourself, then use the multi-part
+-- @crypto_generichash_*()@ APIs and sign the 512-bit output.
```
kozross, line 184: Use the Haskell equivalent - referring to C functions in this way in Haddocks makes very little sense.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,276 @@
...
+
+-- | Opaque tag representing the hash state struct @crypto_sign_state@ used by the C API.
+--
+-- To use a 'CryptoGenericSignState', use 'withGenericSignState'.
```
kozross, line 188: Was this supposed to be `withSignState`?

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,276 @@
...
+--
+-- @since 0.0.1.0
+data CryptoSignState
+
+withSignState :: (Ptr CryptoSignState -> IO a) -> IO a
```
kozross, line 193: You _definitely_ need to document that the `a` should not 'leak' the `Ptr CryptoSignState`, much as `withHandle` should not leak the `Handle`.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,276 @@
...
+withSignState action = do
+  let size = fromIntegral cryptoSignStateBytes
+  allocaBytes size action
+
+-- | The amount of memory needed to store a 'CryptoSignState'.
+foreign import capi "sodium.h crypto_sign_statebytes"
+  cryptoSignStateBytes :: CSize
+
+foreign import capi "sodium.h crypto_sign_init"
```
kozross, line 202: Same issues as for functions in  `Cryptography.Sodium.Bindings.SHA512` apply here.

kozross (changes requested):
The documentation is far from standalone, and in many places does not follow the same conventions as other modules.

src/Cryptography/Sodium/Bindings/Signing.hs
Kleidukos: I'd like to rely on the next iteration of `fourmolu` to unify the comment styles. Is it okay if we do not touch this area until automated tooling does it for us?

kozross: I'm not sure how this is a `fourmolu` concern. My problem is with content, not formatting.

src/Cryptography/Sodium/Bindings/SHA512.hs
```diff
@@ -0,0 +1,122 @@
...
+    -- | @inlen@ parameter. This is the length of the data you want to hash.
+    CULLong ->
+    -- | Return code is 0 on success, -1 on error.
+    IO CInt
+
+-- | This is the state held and used by the various functions of this
+-- module.
+--
+-- @since 0.0.1.0
+data CryptoHashSHA512State = CryptoHashSHA512State
```
kozross, line 65: If we're going for an 'owning' representation like this, it needs a `Storable` instance.

Furthermore, how do you know this is the structure? It's an opaque type as far as I can tell from the docs.

src/Cryptography/Sodium/Bindings/SHA512.hs
```diff
@@ -0,0 +1,122 @@
...
+-- streaming API is also available to process a message as a sequence of multiple chunks.
+
+
+-- | Hash the content of the @in@ buffer and put the result in the @out@ buffer.
+-- The @inlen@ parameter is the size of what is in the @in@ buffer.
+--
+-- @since 0.0.1.0
+foreign import capi "sodium.h crypto_hash_sha512"
+  cryptoHashSHA512 ::
+    -- | @out@ parameter. This is a pointer to the hash of your data.
```
kozross, line 52: I don't think the C parameter names need to be in the docs. Users of the bindings want to know what the parameters are for, not what another language they're deliberately not interacting with chooses to name them.

src/Cryptography/Sodium/Bindings/SHA512.hs
```diff
@@ -0,0 +1,122 @@
...
+-- @since 0.0.1.0
+data CryptoHashSHA512State = CryptoHashSHA512State
+  { state :: Word64,
+    count :: Word64,
+    buf :: Word8
+  }
+
+-- | This function initializes the 'CryptoHashSHA512State' state.
+--
+-- It must imperatively be called before the first 'cryptoHashSHA512Update' call.
```
kozross, line 73: 'imperatively' isn't necessary here.

kozross (changes requested):

src/Cryptography/Sodium/Bindings/Signing.hs
Kleidukos: Ah I see, I somehow mis-read twice your comment. OK!

src/Cryptography/Sodium/Bindings/SHA512.hs
Kleidukos: Ok!

src/Cryptography/Sodium/Bindings/SHA512.hs
Kleidukos: ok!

src/Cryptography/Sodium/Bindings/SHA512.hs
```diff
@@ -0,0 +1,119 @@
...
+import Foreign (Ptr)
+import Foreign.C (CInt (..), CSize (..), CUChar (..), CULLong (..))
+
+-- $introduction
+--
+-- The SHA-512 functions are provided for interoperability with other applications. If you are
+-- looking for a generic hash function and not specifically SHA-2, using
+-- 'Cryptography.Sodium.Bindings.GenericHashing' (BLAKE2b) might be a better choice.
+-- These functions are also not suitable for hashing passwords or deriving keys from passwords.
+-- Use 'Cryptography.Sodium.Bindings.PasswordHashing' instead.
```
kozross, line 39: Would be better phrased as "If this is what you need, use 'Cryptography.Sodium.Bindings.PasswordHashing' instead.".

src/Cryptography/Sodium/Bindings/SHA512.hs
kozross: Empty data type is probably what you want here. We also need a constant defining what size this is, or people won't be able to allocate it.

src/Cryptography/Sodium/Bindings/SHA512.hs
```diff
@@ -0,0 +1,119 @@
...
+
+-- | This is the opaque state held and used by the various functions of this
+-- module.
+--
+-- @since 0.0.1.0
+data CryptoHashSHA512State = CryptoHashSHA512State
+
+-- | This function initializes the 'CryptoHashSHA512State' state.
+--
+-- It must be called before the first 'cryptoHashSHA512Update' call.
```
kozross, line 68: I would phrase it as "Call this function on a 'Ptr CryptoHashSHA512State' before using it as an argument in any other function in this module.".

src/Cryptography/Sodium/Bindings/SHA512.hs
```diff
@@ -0,0 +1,119 @@
...
+
+-- | Add a new chunk to the message that will eventually be hashed.
+--
+-- After all parts have been supplied, 'cryptoHashSHA512Final' can be used to finalise the operation
+-- and get the final hash.
+--
+-- @since 0.0.1.0
+foreign import capi "sodium.h crypto_hash_sha512_update"
+  cryptoHashSHA512Update ::
+    -- | A pointer to the 'CryptoHashSHA512State'. Cannot be 'Foreign.nullPtr'.
```
kozross, line 86: I would rephrase this as "A pointer to an initialized hash state.". As current, you're restating what the type says.

src/Cryptography/Sodium/Bindings/SHA512.hs
```diff
@@ -0,0 +1,119 @@
...
+data CryptoHashSHA512State = CryptoHashSHA512State
+
+-- | This function initializes the 'CryptoHashSHA512State' state.
+--
+-- It must be called before the first 'cryptoHashSHA512Update' call.
+--
+-- @since 0.0.1.0
+foreign import capi "sodium.h crypto_hash_sha512_init"
+  cryptoHashSHA512Init ::
+    -- | A pointer to the 'CryptoHashSHA512State'. Cannot be 'Foreign.nullPtr'.
```
kozross, line 73: It would be better to specify that this should point to a specific number of bytes (as defined by the size constant we need to add) - the current documentation just re-states the type.

src/Cryptography/Sodium/Bindings/SHA512.hs
```diff
@@ -0,0 +1,119 @@
...
+-- | Finalise the hashing of a message. The final hash is padded with extra zeros if necessary,
+-- then put in a buffer.
+--
+-- After this operation, the buffer containing the 'CryptoHashSHA512State' is emptied and
+-- cannot be relied upon.
+--
+-- @since 0.0.1.0
+foreign import capi "sodium.h crypto_hash_sha512_final"
+  cryptoHashSHA512Final ::
+    -- | A pointer to the 'CryptoHashSHA512State'. Cannot be 'Foreign.nullPtr'.
```
kozross, line 104: Same here as for `cryptoHashSHA512Update`.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,328 @@
...
+    -- * Constants
+    cryptoSignPublicKeyBytes,
+    cryptoSignSecretKeyBytes,
+    cryptoSignBytes,
+    cryptoSignSeedBytes,
+  )
+where
+
+import Foreign (Ptr, allocaBytes)
+import Foreign.C (CInt (..), CSize (..), CUChar (..), CULLong (..))
```
kozross, line 48: I am really _not_ a fan of wildcard imports, even for newtypes.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,328 @@
...
+-- Key pair generation --
+-------------------------
+
+-- | This function randomly generates a secret key and a corresponding public key.
+-- The public key is put into the @public key@ and the secret key into the @secret key@.
+--
+-- @since 0.0.1.0
+foreign import capi "sodium.h crypto_sign_keypair"
+  cryptoSignKeyPair ::
+    -- | Pointer to the @public key@. It has a length of 'cryptoSignPublicKeyBytes'.
```
kozross, line 74: I would add 'bytes' to the end of that sentence just to be 100% clear.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,328 @@
...
+    Ptr CUChar ->
+    -- | Pointer to the @secret key@. It has a length of 'cryptoSignSecretKeyBytes'.
```
kozross, line 76: Also here.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,328 @@
...
+-- The signed message, which includes the signature plus an unaltered copy of the message, is put
+-- into @signed message@ and is 'cryptoSignBytes' + @mlen@ bytes long.
+--
+-- If @smlen@ is not a 'Foreign.nullPtr', then the actual length of the signed message is stored in
+-- @smlen@.
+--
+-- @since 0.0.1.0
+foreign import capi "sodium.h crypto_sign"
+  cryptoSign ::
+    -- | Pointer to the @signed message@.
```
kozross, line 112: Why is 'signed message' monospaced here?

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,328 @@
...
+    Ptr CUChar ->
+    -- | Pointer to the length of the @signed message@.
+    Ptr CULLong ->
+    -- | Pointer to the @message@ to sign.
```
kozross, line 116: Why is message monospaced?

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,328 @@
...
+-- Combined Mode --
+-------------------
+
+-- | The 'cryptoSign' function prepends a signature to a message @m@, whose length is @mlen@ bytes,
+-- using the secret key @secret key@.
+--
+-- The signed message, which includes the signature plus an unaltered copy of the message, is put
+-- into @signed message@ and is 'cryptoSignBytes' + @mlen@ bytes long.
+--
+-- If @smlen@ is not a 'Foreign.nullPtr', then the actual length of the signed message is stored in
```
kozross, line 106: What does 'smlen' refer to?

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,328 @@
...
+-- using the secret key @secret key@ and puts the signature into @sig@, which can be up to
+-- 'cryptoSignBytes' bytes long.
+-- The actual length of the signature is put into @siglen@ if @siglen@ is not 'Foreign.nullPtr'.
+-- It is safe to ignore @siglen@ and always consider a signature as 'cryptoSignBytes' bytes long;
+-- shorter signatures will be transparently padded with zeros if necessary.
+--
+-- @since 0.0.1.0
+foreign import capi "sodium.h crypto_sign_detached"
+  cryptoSignDetached ::
+    -- | Pointer to the @signature@.
```
kozross, line 162: Why is 'signature' monospaced?

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,328 @@
...
+  cryptoSignStateBytes :: CSize
+
+-- | This function initializes the 'CryptoSignState' state.
+--
+-- It must be called before the first 'cryptoSignUpdate' call.
+--
+-- @since 0.0.1.0
+foreign import capi "sodium.h crypto_sign_init"
+  cryptoSignInit ::
+    -- | Pointer to the cryptographic state. Cannot be 'Foreign.nullPtr'.
```
kozross, line 227: Similar 'restatement' issues here, as well as needing a more explicit indication of size, as with SHA512.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -0,0 +1,328 @@
...
+    Ptr CryptoSignState ->
+    -- | Return code is 0 on success, -1 on error.
+    IO CInt
+
+-- | Add a new chunk @m@ of length @mlen@ bytes
```
kozross, line 232: Why define `m` and `mlen` here? You never use them again.

kozross (changes requested):
Two things remain here:

* Other function docs in this library have 'linkbacks' to the relevant C API documentation. For consistency, we should have them too.
* The style is a bit non-uniform. Sometimes, it's a high-level description, other times it's a rephrasing of the C docs. I'm much more in favour of the first (see any documentation I've written as an example), but if you prefer a low-level restatement, use it consistently everywhere.

src/Cryptography/Sodium/Bindings/Signing.hs
Kleidukos: I thought it would emphasise the subject? But it can be removed. :)

src/Cryptography/Sodium/Bindings/SHA512.hs
Kleidukos: The libsodium docs say that the size can be retrieved with `cryptoSignStateBytes` (`crypto_sign_statebytes`). If I take a look at the C code, I can instead import a constant, but I think you expressed a preference for the stuff that explicitly appears in the docs, right?

kozross: Yeah, this is definitely the preferred option. Where is `crypto_sign_statebytes()` described in the docs?

src/Cryptography/Sodium/Bindings/Signing.hs
kozross: I think it's better: being monospaced like that suggests a named parameter or a code snippet.

src/Cryptography/Sodium/Bindings/SHA512.hs
Kleidukos: my bad I meant to say `crypto_hash_sha256_statebytes()` (in `crypto_hash_sha256.h`) and `crypto_hash_sha512_statebytes()` (in `crypto_hash_sha512.h`).

Kleidukos: But they do not appear in the documentation actually, only `crypto_sign_statebytes()` for ed25519 is mentioned.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -65,30 +68,36 @@ import Foreign.C (CInt (..), CSize (..), CUChar (..), CULLong (..))
 -- Key pair generation --
 -------------------------
 
--- | This function randomly generates a secret key and a corresponding public key.
--- The public key is put into the @public key@ and the secret key into the @secret key@.
+-- | Randomly generate a secret key and a corresponding public key.
+--
+-- You can retrieve them by accessing the buffers passed as parameters
+-- to this function.
```
kozross, line 74: This is a tad redundant.

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -65,30 +68,36 @@ import Foreign.C (CInt (..), CSize (..), CUChar (..), CULLong (..))
...
-    -- | A pointer to an initialized cryptographic state. It has a length of 'cryptoSignPublicKeyBytes'.
+    -- | A pointer to the buffer holding the public key. It has a length of 'cryptoSignPublicKeyBytes'.
     Ptr CUChar ->
-    -- | A pointer to an initialized cryptographic state. It has a length of 'cryptoSignSecretKeyBytes'.
+    -- | A pointer to the buffer holding the secret key. It has a length of 'cryptoSignSecretKeyBytes'.
     Ptr CUChar ->
-    -- | A pointer to an initialized cryptographic state. It has a length of 'cryptoSignSeedBytes'.
+    -- | A pointer to the seed. It has a length of 'cryptoSignSeedBytes'.
     Ptr CUChar ->
     -- | Return code is 0 on success, -1 on error.
```
kozross, line 102: I would phrase it as "Returns 0 on success, -1 on error.", or even "0 on success, -1 on error.".

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -97,49 +106,52 @@ foreign import capi "sodium.h crypto_sign_seed_keypair"
...
     IO CInt
 
--- | The 'cryptoSignOpen' function checks that the signed message @signed message@,
--- whose length is @smlen@ bytes, has a valid signature for the public key @public key@.
+-- | Check that the signed message has a valid signature for the public key.
 --
--- On success, it puts the message without the signature into @m@, stores its length in @mlen@
--- if @mlen@ is not a 'Foreign.nullPtr' pointer.
+-- On success, it puts the message without the signature into, and stores its length in
+-- the buffer holding the length of the message, if the pointer is not a 'Foreign.nullPtr'.
```
kozross, line 138: This sentence is a bit unclear. It puts the message without the signature into what?

src/Cryptography/Sodium/Bindings/Signing.hs
```diff
@@ -208,23 +245,19 @@ data CryptoSignState
 -- @since 0.0.1.0
 withSignState :: (Ptr CryptoSignState -> IO a) -> IO a
 withSignState action = do
-  size <- fromIntegral <$> cryptoSignStateBytes
+  let size = (fromIntegral @CSize @Int) cryptoSignStateBytes
```
kozross, line 248: You can also spell this as `let size :: Int = fromIntegral cryptoSignStateBytes`.

kozross (changes requested):

src/Cryptography/Sodium/Bindings/SHA2.hs
```diff
@@ -0,0 +1,275 @@
...
+-- | This is the opaque state held and used by the SHA-256 functions.
+--
+-- Its size is 'cryptoHashSHA256StateBytes'.
+--
+-- /See also:/ [crypto_hash_sha256_state](https://doc.libsodium.org/advanced/sha-2_hash_function#data-types)
+--
+-- @since 0.0.1.0
+data CryptoHashSHA256State
+
+-- | Perform an operation with a 'CryptoHashSHA256State of size 'cryptoHashSHA256StateBytes'
```
kozross, line 90: Missing closing quote.

kozross (changes requested):
We should also take care to use `nullPtr` instead of `NULL` when talking about a null pointer in a Haskell context.

src/Cryptography/Sodium/Bindings/SHA2.hs
Kleidukos: wonderful

PR #23, opened 2022-03-29
[WIP] Build from source on Windows

kozross:
This stops bundling binaries on Windows, instead electing to build from source, just like how we do on sensible platforms.


kozross:
Closing this as too impractical.

Issue #15, opened 2022-01-24
Expose constant-time serialisation primitives from libsodium


Kleidukos:
We should expose the constant-time serialisation primitives from libsodium, for hexadecimal and base64.

## Is blocked by

* [x] https://gitlab.haskell.org/ghc/ghc/-/issues/15531 -- Now available since GHC 9.2.2

kpadmasola:
@Kleidukos I'd like to contribute to this to improve my Haskell knowledge. I am a beginner in Haskell, but have started doing a few minor contributions to other Haskell repos. 

Can you please share some pointers on how I should get started with fixing this issue? 

Is there a Discord channel where I can reachout to the maintainers of this codebase for questions?

Thank you!

Kleidukos:
fixed by #33

PR #33, opened 2022-11-12
[#15] Bindings for Hex and Base64 encoding/decoding.

sethlivy:
This PR adds bindings to libsodium's serialization utilities along with the constant-time comparison function.

I've built this on my own machine, but I've never written C bindings before, so I don't have a clue if I'm doing it right.

The associated GHC issue that had been blocking this has been closed and backported to recent GHC versions. (at least from what I read)

### DIFF
--- a/src/LibSodium/Bindings/Utils.hs
+++ b/src/LibSodium/Bindings/Utils.hs
@@ -1,0 +1,155 @@
+{-# LANGUAGE CApiFFI #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE Trustworthy #-}
+{-# OPTIONS_GHC -Wno-unused-imports #-}
+
+-- \| These are bindings to some of sodium's utils.h.
+-- Included are Hex and Base64 encoding/decoding functions
+-- along with a constant-time `memcmp` for handling secret data.
+
+-- | Module: LibSodium.Bindings.Utils
+-- Description: Helpers defined in utils.h
+-- Copyright: (C) Seth Livy 2022
+-- License: BSD-3-Clause
+-- Maintainer: sethlivy@gmail.com
+-- Stability: Stable
+-- Portability: GHC only
+module LibSodium.Bindings.Utils
+  ( -- * Low-level binding
+    sodiumMemcmp
+  , sodiumBin2Hex
+  , sodiumHex2Bin
+  , sodiumBin2Base64
+  , sodiumBase642Bin
+
+    -- * Constants
+  , sodiumBase64VariantOriginal
+  , sodiumBase64VariantOriginalNoPadding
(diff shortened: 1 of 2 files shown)


src/LibSodium/Bindings/Utils.hs
```diff
@@ -0,0 +1,152 @@
...
+-- | Constant-time comparison function.
+--
+-- This function is not a lexicographic comparator and should be never
+-- used for this purpose. It should only be used when comparing two pieces
+-- of secret data, such as keys or authentication tags.
+foreign import capi "sodium.h sodium_memcmp"
+  sodiumMemcmp
+    :: Ptr ()
+    -- ^ First pointer to some secret data.
+    -> Ptr ()
```
kozross, line 43: `()` is _not_ the same as C's `void`. If anything, this should probably be `Ptr CUChar`.

kozross (changes requested):

src/LibSodium/Bindings/Utils.hs
sethlivy: I've changed both of these arguments to `Ptr CUChar`.

kozross (approved):

Kleidukos (commented):
Congratulations, you can now fix the C type warnings! :D https://github.com/haskell-cryptography/libsodium-bindings/actions/runs/3453287531/jobs/5763786083#step:8:1

sethlivy:
Taken a longer look at this. With the C FFI that Haskell provides, we can't represent nested const pointers without dropping the const qualifier and creating errors. 

There is a fix coming for this in GHC 9.6: [https://gitlab.haskell.org/ghc/ghc/-/commit/4f70a8a0b5db49ff249271faefec14bf1421f365](Here.)

src/LibSodium/Bindings/Utils.hs
```diff
@@ -0,0 +1,172 @@
...
+  sodiumBase64VariantOriginal :: CInt
+
+foreign import capi "sodium.h value sodium_base64_VARIANT_ORIGINAL_NO_PADDING"
+  sodiumBase64VariantOriginalNoPadding :: CInt
+
+foreign import capi "sodium.h value sodium_base64_VARIANT_URLSAFE"
+  sodiumBase64VariantURLSafe :: CInt
+
+foreign import capi "sodium.h value sodium_base64_VARIANT_URLSAFE_NO_PADDING"
+  sodiumBase64VariantURLSafeNoPadding :: CInt
```
kozross, line 172: These could do with docs I think.

kozross (changes requested):

kozross (approved):

Kleidukos:
Fantastic, thank you @sethlivy!

PR #34, opened 2022-11-12
Add tests

Kleidukos:
GHC 9.4 support blocked by https://github.com/emilypi/Base16/issues/21

### DIFF
--- a/test/Test/GenericHashing.hs
+++ b/test/Test/GenericHashing.hs
@@ -4,29 +4,25 @@ module Test.GenericHashing where
 
 import Test.Tasty
 import Test.Tasty.HUnit
-import qualified Data.ByteString as BS
 import Foreign hiding (void)
 import Control.Monad (void)
 import Foreign.C
-import LibSodium.Bindings.GenericHashing (cryptoGenericHash, cryptoGenericHashBytes)
+import qualified Data.ByteString.Internal as BS
 import Data.ByteString (ByteString)
-
-useByteString :: ByteString -> ((Ptr a, CSize) -> IO b) -> IO b
-useByteString bs f =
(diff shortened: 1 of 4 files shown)


test/Test/GenericHashing.hs
```diff
@@ -0,0 +1,55 @@
...
+    withCStringLen msg $ \(cString, cstringLength) ->
+    allocaBytes (fromIntegral cryptoGenericHashBytes) $ \outPtr ->
+    BS.unsafeUseAsCStringLen key $ \(keyPtr, keyLength) -> do
+      void $ cryptoGenericHash
+        outPtr
+        cryptoGenericHashBytes
+        (castPtr cString :: Ptr CUChar)
+        (fromIntegral cstringLength)
+        (castPtr keyPtr)
+        (fromIntegral keyLength)
```
sethlivy, line 51: Do we have any other way to convert besides `fromIntegral`? I remember there was some apprehension to using it with C type conversion, but I forget where. 

If this is a non-issue, then it's fine.

sethlivy (approved):

test/Test/GenericHashing.hs
kozross: In this case, it really is a non-issue. `CSize` is equivalent to C's `size_t`; the only guarantees are that it's an unsigned whole-number type at least 16 bits wide. The `Int` that you get from `unsafeUseAsCStringLen` is guaranteed at least 22 bits in size, and GHC Haskell ensures it matches a machine integer on the target platform, which is _at least_ 32 bits, if not 64. Arguably there could be a mismatch, but that's not a problem we can prevent, since it's _entirely_ C-side-dependent.

Kleidukos:
Closed in favour of #35

PR #35, opened 2022-11-16
Add the Sel package

Kleidukos:
I am limiting this PR to Signing & Hashing (Generic & Password) on the `sel` side.

`Safe` markers have been removed.


sel/src/Sel/GenericHashing.hs
```diff
@@ -0,0 +1,98 @@
...
+import Foreign.C (CChar, CSize, CUChar)
+import GHC.IO.Handle.Text (memcpy)
+import LibSodium.Bindings.GenericHashing (cryptoGenericHash, cryptoGenericHashBytes, cryptoGenericHashKeyBytes, cryptoGenericHashKeyGen)
+import System.IO.Unsafe (unsafeDupablePerformIO)
+
+-- |
+--
+-- @since 0.0.1.0
+newtype HashKey = HashKey (ForeignPtr CUChar)
+  deriving newtype (Eq, Ord, Show)
```
kozross, line 29: Need `since`s.

sel/src/Sel/GenericHashing.hs
```diff
@@ -0,0 +1,98 @@
...
+  fPtr <- mallocForeignPtrBytes (fromIntegral cryptoGenericHashKeyBytes)
+  withForeignPtr fPtr $ \ptr ->
+    cryptoGenericHashKeyGen ptr
+  pure $ HashKey fPtr
+
+-- |
+--
+-- @since 0.0.1.0
+newtype Hash = Hash (ForeignPtr CUChar)
+  deriving newtype (Eq, Ord, Show)
```
kozross, line 45: Same here. We also probably want `Storable` at least.

kozross (changes requested):

sel/src/Sel/Hashing/Generic.hs
```diff
@@ -0,0 +1,171 @@
...
+      -- ^ @since 0.0.1.0
+    , Ord
+      -- ^ @since 0.0.1.0
+    , Show
+      -- ^ @since 0.0.1.0
+    )
+
+instance Storable Hash where
+  sizeOf :: Hash -> Int
+  sizeOf _ = sizeOf (undefined :: CUChar)
```
kozross, line 101: Are hashes one byte wide?

sel/src/Sel/Hashing/Generic.hs
```diff
@@ -0,0 +1,171 @@
...
+
+  alignment :: Hash -> Int
+  alignment _ = alignment (undefined :: CUChar)
```
kozross, line 104: This also seems wrong, since that implies you can align on a 1-byte boundary, meaning 'lol wherever'.

sel/src/Sel/Hashing/Generic.hs
```diff
@@ -0,0 +1,171 @@
...
+--
+-- @since 0.0.1.0
+hashToText :: Hash -> Text
+hashToText = Base16.encodeBase16 . hashToBinary
+
+-- | Convert a 'Hash' to a strict, hexadecimal-encoded 'ByteString'.
+--
+-- @since 0.0.1.0
+hashToByteString :: Hash -> ByteString
+hashToByteString = Base16.encodeBase16' . hashToBinary
```
kozross, line 160: I would name these `hashToHexText` and `hashToHexByteString` respectively.

sel/src/Sel/Hashing/Password.hs
```diff
@@ -0,0 +1,206 @@
...
+-- @since 0.0.1.0
+newtype PasswordHash = PasswordHash (ForeignPtr CChar)
+  deriving newtype
+    ( Eq
+      -- ^ @since 0.0.1.0
+    , Ord
+      -- ^ @since 0.0.1.0
+    , Show
+      -- ^ @since 0.0.1.0
+    )
```
kozross, line 74: Same issue with these derivations as before.

sel/src/Sel/Hashing/Password.hs
```diff
@@ -0,0 +1,206 @@
...
+
+-- | @since 0.0.1.0
+instance Display PasswordHash where
+  displayBuilder = Builder.fromText . passwordHashToText
+
+-- | Hash the password with the Argon2id algorithm and
+-- a set of pre-defined parameters
+--
+-- @since 0.0.1.0
+hashPassword :: Text -> IO PasswordHash
```
kozross, line 84: I'd rather accept `ByteString` here, with possibly a `Text` helper. Your data encoding is not our concern.

sel/src/Sel/Hashing/Password.hs
```diff
@@ -0,0 +1,206 @@
...
+          cryptoPWHashOpsLimitModerate
+          cryptoPWHashMemLimitModerate
+    pure $ PasswordHash hashForeignPtr
+
+-- | Verify the password hash.
+--
+-- This function purposefully takes some time to complete, in order to alleviate bruteforce attacks.
+--
+-- @since 0.0.1.0
+verifyPassword :: PasswordHash -> Text -> Bool
```
kozross, line 103: Also here: you shouldn't need to decode data to verify it, it's just bytes.

sel/src/Sel/Signing.hs
```diff
@@ -0,0 +1,210 @@
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+
+-- |
+--
+-- Module: Sel.Hashing.Password
+-- Description: Public-key signatures with the Ed25519 algorithm
```
kozross, line 10: Documentation mismatch?

kozross (changes requested):

sel/src/Sel/Hashing/Generic.hs
```diff
@@ -0,0 +1,171 @@
...
+-- @since 0.0.1.0
+newtype HashKey = HashKey (ForeignPtr CUChar)
+  deriving newtype
+    ( Eq
+      -- ^ @since 0.0.1.0
+    , Ord
+      -- ^ @since 0.0.1.0
+    , Show
+      -- ^ @since 0.0.1.0
+    )
```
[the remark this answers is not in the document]
Kleidukos: @kozross You do not recommend the derivations in principle, or as they stand now and should be properly implemented?

sel/src/Sel/Hashing/Generic.hs
Kleidukos: They can be at most `cryptoGenericHashBytesMax` (64), so I guess I could use CSize? But I don't see a reason why I shouldn't and this bugs me off.

sel/src/Sel/Hashing/Generic.hs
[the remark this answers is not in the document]
kozross: I don't recommend them in principle, because they're almost _never_ what you want. In this specific case, they're not what you want instead.

Basically, implement these to do what you actually want here, not compare pointers.

sel/src/Sel/Hashing/Generic.hs
kozross: So is `Hash` variable-width then?

sel/src/Sel/Hashing/Generic.hs
Kleidukos: Libsodium does some aligning of its own, with compiler internals: [jedisct1/libsodium@ee00928/src/libsodium/include/sodium/export.h#L46-L52](https://github.com/jedisct1/libsodium/blob/ee009287afd480b60a9a4d75b19afdc0f705e9d9/src/libsodium/include/sodium/export.h#L46-L52).

Should we trust it and use 1?

kozross: I'd align on the size of the hash in bytes. A little conservative, but the safest.

Kleidukos: Alright, 32 it is!

kozross (approved):
A few useful instances (such as `Eq`) are missing here, but those could be added later.

Issue #42, opened 2023-02-12
Add Eq & Ord instances to the ForeignPtr newtypes


Kleidukos:
Eq and Ord instances that dereference the `ForeignPtr` should be added

PR #44, opened 2023-02-14
Add Eq & Ord instances to ForeignPtr newtypes. (#42)

sethpaulhubbard:
This PR adds Eq and Ord instances for every newtype that wraps around a `ForeignPtr`. These instances convert the foreign pointer to a regular pointer just like the instances in base, then work on the pointer's contents instead of the address or pointer itself.

---

closes https://github.com/haskell-cryptography/libsodium-bindings/issues/42

### DIFF
--- a/sel/src/Sel/Hashing/Generic.hs
+++ b/sel/src/Sel/Hashing/Generic.hs
@@ -33,12 +33,13 @@ import Data.ByteString.Unsafe (unsafeUseAsCStringLen)
 import Data.Text (Text)
 import Data.Text.Display
 import qualified Data.Text.Lazy.Builder as Builder
-import Foreign (ForeignPtr, Ptr)
+import Foreign (Ptr)
 import qualified Foreign
-import Foreign.C (CSize, CUChar)
+import Foreign.C (CSize(..), CUChar)
 import Foreign.Storable
-import Foreign.ForeignPtr.Unsafe (unsafeForeignPtrToPtr)
+import Foreign.ForeignPtr
 import LibSodium.Bindings.GenericHashing (cryptoGenericHash, cryptoGenericHashBytes, cryptoGenericHashKeyBytes, cryptoGenericHashKeyGen)
+import Sel.Internal
 import System.IO.Unsafe (unsafeDupablePerformIO)
 
 -- $introduction
@@ -67,22 +68,12 @@ import System.IO.Unsafe (unsafeDupablePerformIO)
 newtype HashKey = HashKey (ForeignPtr CUChar)
 
 instance Eq HashKey where
- (HashKey pk1) == (HashKey pk2) = 
-    let p = unsafeForeignPtrToPtr pk1
-        q = unsafeForeignPtrToPtr pk2
-     in unsafeDupablePerformIO $
-          do p' <- peek p
-             q' <- peek q
-             return (p' == q')
+ (HashKey hk1) == (HashKey hk2) = unsafeDupablePerformIO $  
+   unsafeForeignPtrEq hk1 hk2 cryptoGenericHashKeyBytes
 
 instance Ord HashKey where
-  compare (HashKey pk1) (HashKey pk2) =
-    let p = unsafeForeignPtrToPtr pk1
-        q = unsafeForeignPtrToPtr pk2
-     in unsafeDupablePerformIO $
-          do p' <- peek p
-             q' <- peek q
-             return $ compare p' q'
+  compare (HashKey hk1) (HashKey hk2) = unsafeDupablePerformIO $
+    unsafeForeignPtrOrd hk1 hk2 cryptoGenericHashKeyBytes
 
 -- | Create a new 'HashKey' of size 'cryptoGenericHashKeyBytes'.
 --
@@ -105,22 +96,12 @@ newHashKey = do
 newtype Hash = Hash (ForeignPtr CUChar)
 
 instance Eq Hash where
- (Hash pk1) == (Hash pk2) = 
-    let p = unsafeForeignPtrToPtr pk1
-        q = unsafeForeignPtrToPtr pk2
-     in unsafeDupablePerformIO $
-          do p' <- peek p
-             q' <- peek q
-             return (p' == q')
+ (Hash h1) == (Hash h2) = unsafeDupablePerformIO $  
+   unsafeForeignPtrEq h1 h2 cryptoGenericHashBytes
 
 instance Ord Hash where
-  compare (Hash pk1) (Hash pk2) =
-    let p = unsafeForeignPtrToPtr pk1
-        q = unsafeForeignPtrToPtr pk2
-     in unsafeDupablePerformIO $
-          do p' <- peek p
-             q' <- peek q
-             return $ compare p' q'
+  compare (Hash h1) (Hash h2) = unsafeDupablePerformIO $
+   unsafeForeignPtrOrd h1 h2 cryptoGenericHashBytes
 
 instance Storable Hash where
   sizeOf :: Hash -> Int
--- a/sel/src/Sel/Hashing/Generic.hs
+++ b/sel/src/Sel/Hashing/Generic.hs
@@ -37,7 +37,9 @@ import Foreign (ForeignPtr, Ptr)
 import qualified Foreign
 import Foreign.C (CSize, CUChar)
 import Foreign.Storable
+import Foreign.ForeignPtr.Unsafe (unsafeForeignPtrToPtr)
 import LibSodium.Bindings.GenericHashing (cryptoGenericHash, cryptoGenericHashBytes, cryptoGenericHashKeyBytes, cryptoGenericHashKeyGen)
+import System.IO.Unsafe (unsafeDupablePerformIO)
 
 -- $introduction
 --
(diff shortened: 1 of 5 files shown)


sel/src/Sel/Hashing/Generic.hs
```diff
@@ -64,6 +66,24 @@ import LibSodium.Bindings.GenericHashing (cryptoGenericHash, cryptoGenericHashBy
 -- @since 0.0.1.0
 newtype HashKey = HashKey (ForeignPtr CUChar)
 
+instance Eq HashKey where
+ (HashKey pk1) == (HashKey pk2) = 
+    let p = unsafeForeignPtrToPtr pk1
+        q = unsafeForeignPtrToPtr pk2
```
kozross, line 72: This is probably not what you want. You want to use `withForeignPointer` to get at the `Ptr` inside, then compare as before.

sel/src/Sel/Hashing/Generic.hs
```diff
@@ -64,6 +66,24 @@ import LibSodium.Bindings.GenericHashing (cryptoGenericHash, cryptoGenericHashBy
...
+     in unsafeDupablePerformIO $
+          do p' <- peek p
+             q' <- peek q
```
kozross, line 75: This will not work - you are comparing one `CUChar`, but the `HashKey` is longer. You need to use something like `memcmp` or read off in a loop.

sel/src/Sel/Hashing/Generic.hs
```diff
@@ -64,6 +66,24 @@ import LibSodium.Bindings.GenericHashing (cryptoGenericHash, cryptoGenericHashBy
...
+             return (p' == q')
+
+instance Ord HashKey where
+  compare (HashKey pk1) (HashKey pk2) =
+    let p = unsafeForeignPtrToPtr pk1
+        q = unsafeForeignPtrToPtr pk2
+     in unsafeDupablePerformIO $
+          do p' <- peek p
+             q' <- peek q
+             return $ compare p' q'
```
kozross, line 85: Same as above. `memcmp` is _really_ useful here.

sel/src/Sel/Hashing/Generic.hs
```diff
@@ -84,6 +104,24 @@ newHashKey = do
...
+             return (p' == q')
+
+instance Ord Hash where
+  compare (Hash pk1) (Hash pk2) =
+    let p = unsafeForeignPtrToPtr pk1
+        q = unsafeForeignPtrToPtr pk2
+     in unsafeDupablePerformIO $
+          do p' <- peek p
+             q' <- peek q
+             return $ compare p' q'
```
kozross, line 123: Same as for `HashKey`.

sel/src/Sel/Hashing/Password.hs
```diff
@@ -64,6 +65,24 @@ import LibSodium.Bindings.Random
...
+             return (p' == q')
+
+instance Ord PasswordHash where
+  compare (PasswordHash pk1) (PasswordHash pk2) =
+    let p = unsafeForeignPtrToPtr pk1
+        q = unsafeForeignPtrToPtr pk2
+     in unsafeDupablePerformIO $
+          do p' <- peek p
+             q' <- peek q
+             return $ compare p' q'
```
kozross, line 84: Same as for `HashKey`.

sel/src/Sel/Signing.hs
```diff
@@ -73,11 +75,48 @@ import System.IO.Unsafe (unsafeDupablePerformIO)
...
+             return (p' == q')
+
+instance Ord PublicKey where
+  compare (PublicKey pk1) (PublicKey pk2) =
+    let p = unsafeForeignPtrToPtr pk1
+        q = unsafeForeignPtrToPtr pk2
+     in unsafeDupablePerformIO $
+          do p' <- peek p
+             q' <- peek q
+             return $ compare p' q'
```
kozross, line 94: Same as for `HashKey`.

sel/src/Sel/Signing.hs
```diff
@@ -73,11 +75,48 @@ import System.IO.Unsafe (unsafeDupablePerformIO)
...
+             return (p' == q')
+
+instance Ord SecretKey where
+  compare (SecretKey pk1) (SecretKey pk2) =
+    let p = unsafeForeignPtrToPtr pk1
+        q = unsafeForeignPtrToPtr pk2
+     in unsafeDupablePerformIO $
+          do p' <- peek p
+             q' <- peek q
+             return $ compare p' q'
```
kozross, line 118: Same as for `HashKey`.

sel/src/Sel/Signing.hs
```diff
@@ -87,6 +126,32 @@ data SignedMessage = SignedMessage
...
+    let msg1' = unsafeForeignPtrToPtr msg1
+        msg2' = unsafeForeignPtrToPtr msg2
+        sig1' = unsafeForeignPtrToPtr sig1
+        sig2' = unsafeForeignPtrToPtr sig2
+     in unsafeDupablePerformIO $
+          do m1 <- peek msg1'
+             m2 <- peek msg2'
+             s1 <- peek sig1'
+             s2 <- peek sig2'
+             return $ compare (len1, m1, s1) (len2, m2, s2)
```
kozross, line 153: Same as for `HashKey`. Also, the 'repacking' into a tuple is not necessary here; just `&&` for `==` and `<>` for `compare`.

kozross (changes requested):
You need to do these comparisons like you would compare byte buffers in C. Also, `ForeignPtr`s need `withForeignPtr` to be used: `unsafe`-style conversions are not a good idea here.

sel/src/Sel/Hashing/Generic.hs
Kleidukos: Would perhaps https://hackage.haskell.org/package/bytestring-0.11.4.0/docs/Data-ByteString-Internal.html#v:fromForeignPtr work? After all CUChar is a newtype over Word8 right?

Kleidukos: Or would the overhead be too big and indeed, byte-by-byte comparison without creating a bytestring is better?

kozross: Creating a `ByteString` requires _two_ additional `Int`s (length and offset). Thus, this strategy would carry an overhead of 64 bytes, which is a bit ridiculous considering the size of what's being compared.

Seriously, just use `memcmp`, it's not that hard.

sethlivy:
I've resolved all the past conversations since I've redone the entire implementation. This, hopefully, is better than before.

sel/src/Sel/Hashing/Generic.hs
```diff
@@ -33,11 +33,14 @@ import Data.ByteString.Unsafe (unsafeUseAsCStringLen)
 import Data.Text (Text)
 import Data.Text.Display
 import qualified Data.Text.Lazy.Builder as Builder
-import Foreign (ForeignPtr, Ptr)
+import Foreign (Ptr)
 import qualified Foreign
-import Foreign.C (CSize, CUChar)
+import Foreign.C (CSize(..), CUChar)
```
kozross, line 38: Please don't use wildcard imports.

kozross (changes requested):

sel/src/Sel/Internal.hs
```diff
@@ -0,0 +1,29 @@
+{-# LANGUAGE CApiFFI #-}
+{-# LANGUAGE MultiWayIf #-}
+
+module Sel.Internal where
+
+import Foreign.C.Types (CSize(..), CInt(..))
+import Foreign.Ptr (Ptr)
+import Foreign.ForeignPtr (withForeignPtr, ForeignPtr)
+-- | This calls to C's `memcmp` function, used in lieu of
+-- libsodium's `memcmp` in cases when the return code is necessary.
```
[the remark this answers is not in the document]
Kleidukos: I found this in the [helpers](https://libsodium.gitbook.io/doc/helpers) section: 
> Note: sodium_memcmp() is not a lexicographic comparator and is not a generic replacement for memcmp().

sel/src/Sel/Internal.hs
```diff
@@ -0,0 +1,29 @@
...
+
+unsafeForeignPtrOrd :: ForeignPtr a -> ForeignPtr a -> CSize -> IO Ordering
+unsafeForeignPtrOrd fptr1 fptr2 size =
+ withForeignPtr fptr1 $ \p ->
+   withForeignPtr fptr2 $ \q ->
+     do result <- memcmp p q size 
+        return $ if
+          | result == 0 -> EQ
+          | result < 0 -> LT
+          | otherwise -> GT
```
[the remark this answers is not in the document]
sethlivy: Removed the `unsafe` prefix, resolving!

sel/src/Sel/Hashing/Generic.hs
sethlivy: 👍

kozross (approved):

Kleidukos:
@kozross on BSD the last argument is supposed to be a size_t but the error message says it has gotten an unsigned long. Might be related? I'm not super clear on casting rules…

Kleidukos:
No wait, that could be a red herring. How about changing the definition to `string.h memcmp`?

Kleidukos:
Haha yes, perfect.

Issue #41, opened 2023-02-11
Export the AEAD construction XChaCha20-Poly1305


Kleidukos:
https://doc.libsodium.org/secret-key_cryptography/aead/chacha20-poly1305/xchacha20-poly1305_construction

sethpaulhubbard:
I started this a few weeks ago and stopped halfway. Picking this up again now.

Kleidukos:
Closed by #50

PR #50, opened 2023-03-17
(#41) Add AEAD bindings.

sethpaulhubbard:
This PR adds bindings to the XChaCha20-Poly1305 construction in libsodium. Maybe we can do the original and IETF versions later.

### DIFF
--- a/libsodium-bindings/src/LibSodium/Bindings/AEAD.hs
+++ b/libsodium-bindings/src/LibSodium/Bindings/AEAD.hs
@@ -1,0 +1,134 @@
+{-# LANGUAGE CApiFFI #-}
+
+module LibSodium.Bindings.AEAD where
+
+import Foreign.C.Types (CSize(..), CUChar(..), CULLong(..), CInt(..))
+import Foreign.Ptr (Ptr)
+
+-- Bindings to AEAD constructions, specifically XChaCha20.
+
+
+-- | This function encrypts a message, and then appends the authentication tag
+-- to the encrypted message.
+foreign import capi "sodium.h crypto_aead_xchacha20poly1305_ietf_encrypt"
+  cryptoAEADXChaCha20Poly1305IETFEncrypt
+    :: Ptr CUChar
+    -- ^ Output buffer. Contains the encrypted message, authentication tag, and non-confidential additional data.
+    -> Ptr CULLong
+    -- ^ Size of computed output. Should be message length plus crypto_aead_xchacha20poly1305_ietf_ABYTES.
+    -- If set to NULL, then no bytes will be written to this buffer.
+    -> Ptr CUChar
+    -- ^ Message to be encrypted.
+    -> CULLong
+    -- ^ Message length.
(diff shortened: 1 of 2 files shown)


libsodium-bindings/src/LibSodium/Bindings/AEAD.hs
```diff
@@ -0,0 +1,134 @@
+{-# LANGUAGE CApiFFI #-}
+
+module LibSodium.Bindings.AEAD where
```
Kleidukos, line 3: The haddock front-matter is missing, I recommend you just adapt one from the other modules. :)

libsodium-bindings/src/LibSodium/Bindings/AEAD.hs
```diff
@@ -0,0 +1,134 @@
...
+
+import Foreign.C.Types (CSize(..), CUChar(..), CULLong(..), CInt(..))
+import Foreign.Ptr (Ptr)
+
+-- Bindings to AEAD constructions, specifically XChaCha20.
```
Kleidukos, line 8: This would typically be the synopsis in the front-matter

libsodium-bindings/src/LibSodium/Bindings/AEAD.hs
```diff
@@ -0,0 +1,134 @@
...
+
+
+-- | This function encrypts a message, and then appends the authentication tag
+-- to the encrypted message.
```
Kleidukos, line 12: Two things are missing:

* The `@since 0.0.1.0` annotation
* The reference to the libsodium manual, which takes the form of:

```haskell
-- /See:/ [<name of the C function](link to the anchor of the section of the manual)
```

You can generalise this comment to the rest of the functions

libsodium-bindings/src/LibSodium/Bindings/AEAD.hs
```diff
@@ -0,0 +1,134 @@
...
+    -- ^ Length of the additional, non-confidential data.
+    -> Ptr CUChar
+    -- ^ Public nonce of size crypto_aead_xchacha20poly1305_ietf_NPUBBYTES. 
+    -- Should never be reused with the same key. Nonces can be generated using randombytes_buf().
+    -> Ptr CUChar
+    -- ^ Secret key of size crypto_aead_xchacha20poly1305_ietf_KEYBYTES.
+    -> IO CInt
+    -- Returns 0 on success, -1 if tag is not valid.
+
+-- * Constants.
```
Kleidukos, line 122: I'd recommend you split the sections in the explicit export list of the module. :)

Kleidukos (approved):
Good start, some stuff to finish and we should be good :)

Kleidukos (changes requested):
Needs some changes :)

sethpaulhubbard:
Fixed it. Should be good now.

Kleidukos (approved):
Alright, looks good!! :)

PR #66, opened 2023-04-30
Support the ShortHash API

Kleidukos:
Due to the exposed API being quite partial regarding the SipHashX24 variant, I have provided an equivalent of the keygen function for that variant that relies on randombytes_buf. This is how the keygen function is actually implemented, just at the C level instead of Haskell-level.


libsodium-bindings/src/LibSodium/Bindings/ShortHash.hs
```diff
@@ -0,0 +1,138 @@
...
+--
+-- It is implemented with 'randombytesBuf'
+--
+-- @since 0.0.1.0
+cryptoShortHashX24KeyGen
+  :: Ptr CUChar
+  -- ^ Buffer that will hold the secret key.
+  -> IO ()
+cryptoShortHashX24KeyGen ptr =
+  randombytesBuf (coerce ptr :: Ptr Word8) cryptoShortHashSipHashX24KeyBytes
```
kozross, line 84: Why `coerce` instead of `castPtr`?

kozross (changes requested):

libsodium-bindings/src/LibSodium/Bindings/ShortHash.hs
Kleidukos: Because I had forgotten its existence.

kozross (approved):

PR #67, opened 2023-05-07
Implement the Key Derivation API

### LINKED ISSUE #63: Support Key derivation API
https://doc.libsodium.org/key_derivation

Kleidukos:
Fixes #63

### DIFF
--- a/libsodium-bindings/src/LibSodium/Bindings/KeyDerivation.hs
+++ b/libsodium-bindings/src/LibSodium/Bindings/KeyDerivation.hs
@@ -1,0 +1,81 @@
+{-# LANGUAGE CApiFFI #-}
+
+-- |
+-- Module: LibSodium.Bindings.KeyDerivation
+-- Description: Direct bindings to the key exchange functions implemented in Libsodium
+-- Copyright: (C) Hécate Moonlight 2022
+-- License: BSD-3-Clause
+-- Maintainer: The Haskell Cryptography Group
+-- Stability: Stable
+-- Portability: GHC only
+module LibSodium.Bindings.KeyDerivation
+  ( -- * Introduction
+    -- $introduction
+
+    -- ** Key Generation
+    cryptoKDFKeygen
+  , cryptoKDFDeriveFromKey
+
+    -- ** Constants
+  , cryptoKDFPrimitive
+  , cryptoKDFBytesMin
+  , cryptoKDFBytesMax
+  , cryptoKDFKeyBytes
+  , cryptoKDFContextBytes
+  )
+where
+
+import Foreign (Ptr, Word64, Word8)
+import Foreign.C (CChar, CInt (CInt), CSize (CSize), CUChar)
+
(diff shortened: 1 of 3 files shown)


libsodium-bindings/src/LibSodium/Bindings/KeyDerivation.hs
```diff
@@ -0,0 +1,81 @@
...
+    -- ^ Pointer that will hold the sub-key.
+    -> CSize
+    -- ^ Length of the sub-key.
+    -> Word64
+    -- ^ Identifier of the sub-key. Must not be reused for another sub-key.
+    -> Ptr CChar
+    -- ^ Pointer to the context, of size 'cryptoKDFContextBytes'.
+    -> Ptr CUChar
+    -- ^ Pointer to the master key, which will be of length 'cryptoKDFKeyBytes'.
+    -> IO CInt
```
kozross, line 49: What does the return value mean?

libsodium-bindings/src/LibSodium/Bindings/KeyDerivation.hs
```diff
@@ -0,0 +1,81 @@
...
+
+-- == Constants
+
+-- | The algorithm used by the key derivation algorithm, blake2b.
+--
+-- @since 0.0.1.0
+cryptoKDFPrimitive :: String
+cryptoKDFPrimitive = "blake2b"
```
kozross, line 57: Is this something you're meant to pass into C? If so, `String` won't work, you need a C string.

kozross (changes requested):

libsodium-bindings/src/LibSodium/Bindings/KeyDerivation.hs
Kleidukos: No, it's a fixed string on the C side but the C compiler is giving me shit about it.

```c

#define crypto_kdf_PRIMITIVE "blake2b"
SODIUM_EXPORT
const char *crypto_kdf_primitive(void)
            __attribute__ ((warn_unused_result));
```

I tried a lot of export types, and it was never quite right, so I gave up and just re-used the string.

libsodium-bindings/src/LibSodium/Bindings/KeyDerivation.hs
Kleidukos: Ah good catch

libsodium-bindings/src/LibSodium/Bindings/KeyDerivation.hs
kozross: What exact error are you getting? A 'fixed string on the C side' is _nothing whatsoever_ like a Haskell `String`, so whatever the cause, your solution won't work.

kozross: Does `Ptr CChar` not work?

kozross (approved):

PR #76, opened 2023-05-18
SecretKey.AuthenticatedEncryption

Kleidukos:
Here is a documentation screenshot:
image

### DIFF
--- a/sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
+++ b/sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
@@ -48,9 +48,18 @@ import Sel.Internal
 -- Poly1305 MAC hash.
 
 -- $usage
--- TODO
+--
+-- > import qualified Sel.SecretKey.AuthenticatedEncryption as AuthenticatedEncryption
+-- >
+-- > main = do
+-- >   -- We get the secretKey from the other party or with 'newSecretKey'.
+-- >   -- We get the nonce from the other party or with 'newNonce'. Do not reuse it with the same secret key!
+-- >   encryptedMessage <- AuthenticatedEncryption.encrypt "hello hello" secretKey nonce
+-- >   let result = AuthenticatedEncryption.decrypt encryptedMessage secretKey nonce
+-- >   print result
+-- >   -- "Just \"hello hello\""
 
--- |
+-- | A secret key of size 'cryptoSecretboxKeyBytes'.
 --
 -- @since 0.0.1.0
 newtype SecretKey = SecretKey (ForeignPtr CUChar)
@@ -71,7 +80,9 @@ instance Ord SecretKey where
     unsafeDupablePerformIO $
       foreignPtrOrd hk1 hk2 cryptoSecretboxKeyBytes
 
--- |
+-- | A random number that must only be used once per exchanged message.
+-- It does not have to be confidential.
+-- It is of size 'cryptoSecretboxNonceBytes'.
 --
 -- @since 0.0.1.0
 newtype Nonce = Nonce (ForeignPtr CUChar)
@@ -92,7 +103,7 @@ instance Ord Nonce where
     unsafeDupablePerformIO $
       foreignPtrOrd hk1 hk2 cryptoSecretboxKeyBytes
 
--- |
+-- | Generate a new random secret key.
 --
 -- @since 0.0.1.0
 newSecretKey :: IO SecretKey
@@ -102,7 +113,8 @@ newSecretKey = do
     cryptoSecretboxKeygen ptr
   pure $ SecretKey fPtr
 
--- |
+-- | Generate a new random nonce.
+-- Only use it once per exchanged message.
 --
 -- @since 0.0.1.0
 newNonce :: IO Nonce
@@ -112,7 +124,7 @@ newNonce = do
     randombytesBuf (Foreign.castPtr @CUChar @Word8 ptr) cryptoSecretboxNonceBytes
   pure $ Nonce fPtr
 
--- |
+-- | A ciphertext consisting of an encrypted message and an authentication tag.
 --
 -- @since 0.0.1.0
 data Hash = Hash
@@ -140,11 +152,11 @@ instance Ord Hash where
 
 encrypt
   :: StrictByteString
-  -- ^ Message to encrypt
+  -- ^ Message to encrypt.
   -> SecretKey
-  -- ^ Secret key generated with 'newSecretKey'
+  -- ^ Secret key generated with 'newSecretKey'.
   -> Nonce
-  -- ^ One-time use number generated with 'newNonce'
+  -- ^ One-time use number generated with 'newNonce'.
   -> IO Hash
 encrypt message (SecretKey secretKeyForeignPtr) (Nonce nonceForeignPtr) =
   BS.unsafeUseAsCStringLen message $ \(cString, cStringLen) -> do
@@ -163,11 +175,11 @@ encrypt message (SecretKey secretKeyForeignPtr) (Nonce nonceForeignPtr) =
 
 decrypt
   :: Hash
-  -- ^ Encrypted message
+  -- ^ Encrypted message you want to decrypt.
   -> SecretKey
-  -- ^ Secret key used for encrypting the original message
+  -- ^ Secret key used for encrypting the original message.
   -> Nonce
-  -- ^ Nonce used for encrypting the original message
+  -- ^ Nonce used for encrypting the original message.
   -> Maybe StrictByteString
 decrypt Hash{messageLength, hashForeignPtr} (SecretKey secretKeyForeignPtr) (Nonce nonceForeignPtr) = unsafeDupablePerformIO $ do
   messagePtr <- Foreign.mallocBytes (fromIntegral @CULLong @Int messageLength)
--- a/sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
+++ b/sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
@@ -28,7 +28,7 @@ module Sel.SecretKey.AuthenticatedEncryption
     -- ** Nonce
   , Nonce
   , nonceFromByteString
-  , unsafeNonceToHexByteString
+  , nonceToHexByteString
 
     -- ** Hash
   , Hash
@@ -44,11 +44,12 @@ module Sel.SecretKey.AuthenticatedEncryption
 
 import Control.Monad (void)
 import Data.ByteString (StrictByteString)
+import qualified Data.ByteString as BS
 import qualified Data.ByteString.Base16 as Base16
 import qualified Data.ByteString.Internal as BS
 import qualified Data.ByteString.Unsafe as BS
 import Data.Text (Text)
-import Data.Text.Display (Display (displayBuilder), OpaqueInstance (..))
+import Data.Text.Display (Display (displayBuilder), OpaqueInstance (..), ShowInstance (..))
 import qualified Data.Text.Lazy.Builder as Builder
 import Data.Word (Word8)
 import Foreign (ForeignPtr)
@@ -132,9 +133,13 @@ newSecretKey = do
 secretKeyFromByteString :: StrictByteString -> SecretKey
 secretKeyFromByteString bytestring = unsafeDupablePerformIO $
   BS.unsafeUseAsCStringLen bytestring $ \(outsideSecretKeyPtr, _) -> do
-    secretKeyForeignPtr <- BS.mallocByteString @CChar (fromIntegral cryptoSecretboxKeyBytes)
+    secretKeyForeignPtr <-
+      BS.mallocByteString @CChar (fromIntegral cryptoSecretboxKeyBytes)
     Foreign.withForeignPtr secretKeyForeignPtr $ \secretKeyPtr ->
-      Foreign.copyArray outsideSecretKeyPtr secretKeyPtr (fromIntegral cryptoSecretboxKeyBytes)
+      Foreign.copyArray
+        outsideSecretKeyPtr
+        secretKeyPtr
+        (fromIntegral cryptoSecretboxKeyBytes)
     pure $ SecretKey (Foreign.castForeignPtr @CChar @CUChar secretKeyForeignPtr)
 
 -- | Convert a 'SecretKey' to a hexadecimal-encoded 'StrictByteString'.
@@ -160,7 +165,7 @@ newtype Nonce = Nonce (ForeignPtr CUChar)
       -- ^ @since 0.0.1.0
       -- > display secretKey == "[REDACTED]"
     )
-    via (OpaqueInstance "[REDACTED]" Nonce)
+    via (ShowInstance Nonce)
 
 -- |
 --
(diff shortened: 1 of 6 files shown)


sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,208 @@
...
+      Foreign.withForeignPtr secretKeyForeignPtr $ \secretKeyPtr ->
+        Foreign.withForeignPtr nonceForeignPtr $ \noncePtr -> do
+          void $
+            cryptoSecretboxEasy
+              hashPtr
+              (Foreign.castPtr @CChar @CUChar cString)
+              (fromIntegral @Int @CULLong cStringLen)
+              noncePtr
+              secretKeyPtr
+    pure $ Hash (fromIntegral @Int @CULLong cStringLen) hashForeignPtr
```
kozross, line 178: Since nonces are meant to be single-use, it would be better to have:

``encrypt :: StrictByteString -> SecretKey -> IO (Nonce, Hash)``

That way, nonces can never be re-used.

kozross (changes requested):

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
Kleidukos: Yes, very smart. :)

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,208 @@
...
+import System.IO.Unsafe (unsafeDupablePerformIO)
+
+import Control.Monad (void)
+import Data.Word (Word8)
+import LibSodium.Bindings.Random (randombytesBuf)
+import LibSodium.Bindings.Secretbox (cryptoSecretboxEasy, cryptoSecretboxKeyBytes, cryptoSecretboxKeygen, cryptoSecretboxMACBytes, cryptoSecretboxNonceBytes, cryptoSecretboxOpenEasy)
+import Sel.Internal
+
+-- $introduction
+-- Authenticated Encryption is the action of encrypting a message using a secret key
```
a-02, line 42: Incredibly minor suggestion: `"Authenticated Encryption" uses a secret key along with a single-use number called a "nonce" to encrypt a message.`

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,208 @@
...
+-- |
+--
+-- @since 0.0.1.0
+instance Ord Hash where
+  compare (Hash messageLength1 hk1) (Hash messageLength2 hk2) =
+    unsafeDupablePerformIO $ do
+      result1 <- foreignPtrOrd hk1 hk2 (fromIntegral messageLength1 + cryptoSecretboxMACBytes)
+      pure $ compare messageLength1 messageLength2 <> result1
+
+-- | Create an authenticated hash from a message, a secret key that must remain secret, and a one-time cryptographic nonce
```
a-02, line 153: "A secret key that must remain secret" is redundant. The secret key is already secret.

a-02 (approved):

Kleidukos:
Okay, I've added import/export functions, and warnings

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,388 @@
...
+--
+-- @since 0.0.1.0
+instance Display SecretKey where
+  displayBuilder = Builder.fromText . secretKeyToHexText
+
+-- | ⚠️  Be prudent with what you do with it!
+--
+-- @since 0.0.1.0
+instance Show SecretKey where
+  show = BS.unpackChars . secretKeyToHexByteString
```
kozross, line 118: Might be worth explaining why this is somewhat dangerous. I would actually argue that it might be better to have `Show` and `Display` both do censoring, and provide something like `unsafeDisplayNoDon'tDoThisWeMeanIt` to show it in hex.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,388 @@
...
+-- usually from the network or disk.
+--
+-- @since 0.0.1.0
+secretKeyFromByteString :: StrictByteString -> SecretKey
+secretKeyFromByteString bytestring = unsafeDupablePerformIO $
+  BS.unsafeUseAsCStringLen bytestring $ \(outsideSecretKeyPtr, _) -> do
+    secretKeyForeignPtr <- BS.mallocByteString @CChar (fromIntegral cryptoSecretboxKeyBytes)
+    Foreign.withForeignPtr secretKeyForeignPtr $ \secretKeyPtr ->
+      Foreign.copyArray outsideSecretKeyPtr secretKeyPtr (fromIntegral cryptoSecretboxKeyBytes)
+    pure $ SecretKey (Foreign.castForeignPtr @CChar @CUChar secretKeyForeignPtr)
```
kozross, line 140: Will this accept _absolutely any_ `StrictByteString`? I assume there's a length requirement.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,388 @@
...
+secretKeyToHexText :: SecretKey -> Text
+secretKeyToHexText = Base16.encodeBase16 . secretKeyToBinary
+
+-- | Convert a 'SecretKey' to a strict, hexadecimal-encoded 'StrictByteString'.
+--
+-- ⚠️  Be prudent as to where you store it!
+--
+-- @since 0.0.1.0
+secretKeyToHexByteString :: SecretKey -> StrictByteString
+secretKeyToHexByteString = Base16.encodeBase16' . secretKeyToBinary
```
kozross, line 156: Same as before.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,388 @@
...
+-- | Convert a 'SecretKey' to a strict binary 'StrictByteString'.
+--
+-- ⚠️  Be prudent as to where you store it!
+--
+-- @since 0.0.1.0
+secretKeyToBinary :: SecretKey -> StrictByteString
+secretKeyToBinary (SecretKey secretKeyForeignPtr) =
+  BS.fromForeignPtr0
+    (Foreign.castForeignPtr @CUChar @Word8 secretKeyForeignPtr)
+    (fromIntegral @CSize @Int cryptoSecretboxKeyBytes)
```
kozross, line 167: Here too.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,388 @@
...
+--
+-- @since 0.0.1.0
+instance Display Nonce where
+  displayBuilder = Builder.fromText . nonceToHexText
+
+-- | ⚠️  Be prudent with what you do with it!
+--
+-- @since 0.0.1.0
+instance Show Nonce where
+  show = BS.unpackChars . nonceToHexByteString
```
kozross, line 202: Why would this be a problem if it's not confidential?

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,388 @@
...
+-- | Generate a new random nonce.
+-- Only use it once per exchanged message.
+--
+-- Do not use this outside of hash creation!
+newNonce :: IO Nonce
+newNonce = do
+  (fPtr :: ForeignPtr CUChar) <- Foreign.mallocForeignPtrBytes (fromIntegral cryptoSecretboxNonceBytes)
+  Foreign.withForeignPtr fPtr $ \ptr ->
+    randombytesBuf (Foreign.castPtr @CUChar @Word8 ptr) cryptoSecretboxNonceBytes
+  pure $ Nonce fPtr
```
kozross, line 213: Why would you even need this? Nonces can only show up in two ways that make sense:

* Read from the outside world; or
* During the process of encryption.

This helps neither, and acts as a very large footgun.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,388 @@
...
+-- usually from the network or disk.
+--
+-- @since 0.0.1.0
+nonceFromByteString :: StrictByteString -> Nonce
+nonceFromByteString bytestring = unsafeDupablePerformIO $
+  BS.unsafeUseAsCStringLen bytestring $ \(outsideNoncePtr, _) -> do
+    nonceForeignPtr <- BS.mallocByteString @CChar (fromIntegral cryptoSecretboxNonceBytes)
+    Foreign.withForeignPtr nonceForeignPtr $ \noncePtr ->
+      Foreign.copyArray outsideNoncePtr noncePtr (fromIntegral cryptoSecretboxNonceBytes)
+    pure $ Nonce (Foreign.castForeignPtr @CChar @CUChar nonceForeignPtr)
```
kozross, line 225: Will _any_ `StrictByteString` do? I assume there's a size requirement.

For the point of view of 'reading from network or disk', I'd rather have fewer footguns by doing something like:

```
nonceFromHandle :: Handle StrictByteString -> (Maybe Nonce -> IO r) -> IO r
```

with a giant 'do not leak this nonce no we mean it'.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,388 @@
...
+-- | Convert a 'Nonce' to a strict binary 'StrictByteString'.
+--
+-- ⚠️  Be prudent as to where you store it!
+--
+-- @since 0.0.1.0
+nonceToBinary :: Nonce -> StrictByteString
+nonceToBinary (Nonce nonceForeignPtr) =
+  BS.fromForeignPtr0
+    (Foreign.castForeignPtr @CUChar @Word8 nonceForeignPtr)
+    (fromIntegral @CSize @Int cryptoSecretboxKeyBytes)
```
kozross, line 252: If nonces aren't confidential, why is this a problem?

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,388 @@
...
+hashFromByteString :: StrictByteString -> Hash
+hashFromByteString bytestring = unsafeDupablePerformIO $
+  BS.unsafeUseAsCStringLen bytestring $ \(outsideHashPtr, outsideHashLength) -> do
+    hashForeignPtr <- BS.mallocByteString @CChar outsideHashLength
+    Foreign.withForeignPtr hashForeignPtr $ \hashPtr ->
+      Foreign.copyArray outsideHashPtr hashPtr outsideHashLength
+    pure $
+      Hash
+        (fromIntegral @Int @CULLong outsideHashLength)
+        (Foreign.castForeignPtr @CChar @CUChar hashForeignPtr)
```
kozross, line 305: How is it encoded? Will _any_ `StrictByteString` do?

kozross (changes requested):

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
Kleidukos: ah yes thank you

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
Kleidukos: Because caution is the mother of safety, after all they are essential in making sure replay attacks are impossible

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
Kleidukos: It's not exposed, I just need it in the `encrypt` function.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
kozross: Something doesn't add up for me here: on the one hand, our documentation claims that nonces aren't confidential, but on the other, if they can facilitate replays, they sound pretty damn confidential to me.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
kozross: Ah, OK. Then why does it have Haddocks?

Kleidukos: Because it's so damn convenient :) 
image

kozross: I genuinely don't believe 'IDE support' is sufficient reason for Haddocks on a private helper. However, I'm willing to concede that one. I just find it personally confusing.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
Kleidukos: It's the concatenation of the message + authentication tag, then encrypted. I guess the only size constant here is the MAC.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
Kleidukos: You're right, libsodium does say the nonces don't have to be confidential, but I feel quite paranoid on these matters

kozross: Paranoia is fine, as long as it's justified. I guess I'm just not clear on what 'not confidential' means. Probably needs an actual expert to weigh in.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
kozross: That suggests it could fail.

kozross (changes requested):

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -277,20 +245,21 @@ instance Ord Hash where
...
   show = BS.unpackChars . hashToHexByteString
 
--- | Create a 'Hash' from a 'StrictByteString' that you have obtained on your own,
--- usually from the network or disk.
+-- | Create a 'Hash' from a binary 'StrictByteString' that you have obtained on your own,
+-- usually from the network or disk. It must be a valid hash built from the concatenation
+-- of the encrypted message and the authentication tag.
 --
 -- @since 0.0.1.0
 hashFromByteString :: StrictByteString -> Hash
```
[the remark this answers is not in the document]
Kleidukos: I guess the only meaningful check would be to make sure that there is at least `cryptoSecretboxMACBytes`. However I do not see any kind of check in other high-level implementations. For example [alkali](https://docs.rs/alkali/latest/alkali/symmetric/cipher/xsalsa20poly1305/fn.decrypt.html) takes directly an array of u8 (that would be Word8 for us, which is a bytestring).

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
cronokirby: Nonces need to be unique for each encryption, but can be public, and generated in a predictable way.

Kleidukos: Thanks! I can now safely withdraw my fears :)

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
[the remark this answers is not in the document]
kozross: I guess if you decode something that doesn't make sense, it won't decode. I'd rather still do whatever checks are possible.

Kleidukos:
@kozross alright I've added length checks for the bytestrings :)

kozross (approved):

PR #79, opened 2023-05-20
Make usage of `MultiPart` more secure

### LINKED ISSUE #78: Suggestions wrt `MultiPart`
It is said, that it would be bad if the Multipart escapes its scope, so perhaps we can do the skolem trick on it. I don't know exactly whether this is sound, it's merely an idea and perhaps it's non-sense: 

```haskell
data Multipart s = Multipart

withMultipart
  :: MonadIO m
  => (forall s. Multipart s -> m a)
  -> m a
withMultipart k = k Multipart

escapeMultiPart :: IORef (Multipart s) -> IO ()
escapeMultiPart ref = withMultipart (\part ->  writeIORef ref part)
```

MangoIV:
- resolves #78 
- adds a tag `s` to `Multipart` which should make it harder to escape a `Multipart` out of its scope in `withMultipart` 
- before it was possible to just escape the `Multipart` by using `pure` 

> **Warning**
> this should be carefully checked for soundness by reviewers

### DIFF
--- a/sel/src/Sel/Hashing/SHA2/SHA512.hs
+++ b/sel/src/Sel/Hashing/SHA2/SHA512.hs
@@ -1,5 +1,7 @@
 {-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE RoleAnnotations #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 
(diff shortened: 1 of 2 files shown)


sel/src/Sel/Hashing/SHA2/SHA512.hs
```diff
@@ -179,7 +180,7 @@ hashText text = hashByteString (Text.encodeUtf8 text)
 -- ...   SHA512.finaliseMultipart multipartState
 --
 -- @since 0.0.1.0
-newtype Multipart = Multipart (Ptr CryptoHashSHA512State)
+newtype Multipart s = UnsafeMultipart (Ptr CryptoHashSHA512State)
```
Kleidukos, line 183: I don't think you need to rename the data constructor because we're not exporting it. :)

MangoIV: ah yeah, I just thought that even though we're not currently it would be nice to make it clear that usage of this constructor is generally unsafe.

Kleidukos:
@MangoIV just got to run Fourmolu and it should be good!

MangoIV:
dun @Kleidukos

Kleidukos (approved):

PR #82, opened 2023-06-03
Implement Encrypted Streams

Kleidukos:
## TODO

- [x] `Multipart` continuation function
- [x] Usage example at the top
- [ ] push / pull with additional data

image

### DIFF
--- a/sel/src/Sel/SecretKey/EncryptedStream.hs
+++ b/sel/src/Sel/SecretKey/EncryptedStream.hs
@@ -19,7 +19,7 @@ module Sel.SecretKey.EncryptedStream
   ( -- ** Usage
     -- $usage
 
-  -- ** Secret Key
+    -- ** Secret Key
     SecretKey
   , newSecretKey
 
@@ -94,10 +94,6 @@ import LibSodium.Bindings.SecretStream
   , cryptoSecretStreamXChaCha20Poly1305TagRekey
   )
 import Sel.Internal
-
--- $usage
---
---
 
 -- | The 'SecretKey' is used to encrypt the stream.
 --
--- a/sel/src/Sel/SecretKey/EncryptedStream.hs
+++ b/sel/src/Sel/SecretKey/EncryptedStream.hs
@@ -50,17 +50,20 @@ module Sel.SecretKey.EncryptedStream
     -- *** Encryption
   , initPushStream
   , pushToStream
+  , pushToStreamWith
 
     -- *** Decryption
   , StreamResult (..)
   , initPullStream
   , pullFromStream
+  , pullFromStreamWith
 
     -- *** Key regeneration
   , rekey
   ) where
 
 import Control.Monad.IO.Class (MonadIO, liftIO)
+import Control.Monad (void)
 import Data.ByteString (StrictByteString)
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Base16 as Base16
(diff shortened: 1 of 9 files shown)


sel/src/Sel/SecretKey/EncryptedStream.hs
```diff
@@ -0,0 +1,542 @@
...
+    unsafeDupablePerformIO $
+      foreignPtrEq hk1 hk2 cryptoSecretStreamXChaCha20Poly1305KeyBytes
+
+-- |
+--
+-- @since 0.0.1.0
+instance Ord SecretKey where
+  compare (SecretKey hk1) (SecretKey hk2) =
+    unsafeDupablePerformIO $
+      foreignPtrOrd hk1 hk2 cryptoSecretStreamXChaCha20Poly1305KeyBytes
```
kozross, line 119: We do this in a bunch of places. Might be worth writing a helper `newtype` for DerivingVia for these.

sel/src/Sel/SecretKey/EncryptedStream.hs
```diff
@@ -0,0 +1,542 @@
...
+-- If the message is not strictly longer than the size of the authentication tag,
+-- the function will fail.
+--
+-- @since 0.0.1.0
+cipherTextFromByteString :: StrictByteString -> Maybe CipherText
+cipherTextFromByteString bytestring =
+  let (foreignPtr, bsLength) = BS.toForeignPtr0 bytestring
+   in if bsLength > fromIntegral @CSize @Int cryptoSecretStreamXChaCha20Poly1305ABytes
+        then Just $ CipherText (Foreign.castForeignPtr @Word8 @CUChar foreignPtr) (fromIntegral @Int @CSize bsLength)
+        else Nothing
```
kozross, line 160: I'm not sure this is completely safe: we copy a pointer to some foreign data, but not the data itself, which means that when the `ByteString` drops out of scope, our pointer now dangles. An 'owning copy' would be much better.

sel/src/Sel/SecretKey/EncryptedStream.hs
```diff
@@ -0,0 +1,542 @@
...
+
+-- | Convert a 'CipherText' to a binary 'StrictByteString'.
+--
+-- @since 0.0.1.0
+cipherTextToBinary :: CipherText -> StrictByteString
+cipherTextToBinary (CipherText fPtr size) =
+  BS.fromForeignPtr0
+    (Foreign.castForeignPtr fPtr)
+    (fromIntegral size)
```
kozross, line 169: Similar to `cipherTextFromByteString`.

sel/src/Sel/SecretKey/EncryptedStream.hs
```diff
@@ -0,0 +1,542 @@
...
+--
+-- If the bytestring is not of the appropriate length, the function will fail.
+--
+-- @since 0.0.1.0
+headerFromByteString :: StrictByteString -> Maybe Header
+headerFromByteString bytestring =
+  let (foreignPtr, bsLength) = BS.toForeignPtr0 bytestring
+   in if bsLength == fromIntegral @CSize @Int cryptoSecretStreamXChaCha20Poly1305HeaderBytes
+        then Just $ Header (Foreign.castForeignPtr @Word8 @CUChar foreignPtr)
+        else Nothing
```
kozross, line 255: Here as well.

sel/src/Sel/SecretKey/EncryptedStream.hs
```diff
@@ -0,0 +1,542 @@
...
+      -- ^ @since 0.0.1.0
+    , Show
+      -- ^ @since 0.0.1.0
+    , Ord
+      -- ^ @since 0.0.1.0
+    , Enum
+      -- ^ @since 0.0.1.0
+    , Bounded
+      -- ^ @since 0.0.1.0
+    )
```
kozross, line 283: Needs a `stock`.

sel/src/Sel/SecretKey/EncryptedStream.hs
```diff
@@ -0,0 +1,542 @@
...
+  , mAdditionalData :: Maybe StrictByteString
+  }
+  deriving
+    ( Eq
+      -- ^ @since 0.0.1.0
+    , Show
+      -- ^ @since 0.0.1.0
+    , Ord
+      -- ^ @since 0.0.1.0
+    )
```
kozross, line 314: Also here.

sel/src/Sel/SecretKey/EncryptedStream.hs
```diff
@@ -0,0 +1,542 @@
...
+  | InputStreamInitError
+  | InputStreamPushError
+  deriving
+    ( Eq
+      -- ^ @since 0.0.1.0
+    , Show
+      -- ^ @since 0.0.1.0
+    , Ord
+      -- ^ @since 0.0.1.0
+    )
```
kozross, line 336: Here also.

sel/src/Sel/SecretKey/EncryptedStream.hs
```diff
@@ -0,0 +1,542 @@
...
+  Foreign.withForeignPtr headerForeignPtr $ \headerPtr ->
+    Foreign.withForeignPtr secretKeyForeignPtr $ \keyPtr -> do
+      result <-
+        cryptoSecretStreamXChaCha20Poly1305InitPush
+          statePtr
+          headerPtr
+          keyPtr
+      case result of
+        0 -> pure $ Right (Header headerForeignPtr, SecretKey secretKeyForeignPtr)
+        _ -> pure $ Left InputStreamInitError
```
kozross, line 366: This leaks memory, unless `mallocForeignPtrBytes` automagically installs a finalizer.

sel/src/Sel/SecretKey/EncryptedStream.hs
```diff
@@ -0,0 +1,542 @@
...
+              Foreign.nullPtr
+              (Foreign.castPtr @CChar @CUChar cString)
+              (fromIntegral @Int @CULLong cStringLen)
+              additionalDataPointer
+              additionalDataLength
+              (streamTagToConstant tag)
+          case result of
+            0 -> do
+              Foreign.withForeignPtr cipherTextFPtr $ \cipherTextHomePtr ->
+                memcpy cipherTextBuffer cipherTextHomePtr (fromIntegral cipherTextLength)
```
kozross, line 409: This is a bit inefficient: you allocate a temporary buffer, only to then copy it immediately on success.

kozross (changes requested):

sel/src/Sel/SecretKey/EncryptedStream.hs
Kleidukos: Something like this?

```haskell
      cipherForeignPtr <- Foreign.mallocPlainForeignPtrBytes (BS.length bytestring)
      Foreign.withForeignPtr cipherForeignPtr $ \cipherPtr ->
        BS.unsafeUseAsCStringLen bytestring $ \(cString, cStringLen) -> do
          memcpy (Foreign.castPtr cipherPtr) cString (fromIntegral cStringLen)
          pure $
            Just $
              CipherText
                cipherForeignPtr
                (fromIntegral @Int @CSize cStringLen)
```

sel/src/Sel/SecretKey/EncryptedStream.hs
Kleidukos: Actually yes! https://hackage.haskell.org/package/base-4.18.0.0/docs/Foreign-ForeignPtr.html#v:mallocForeignPtr

sel/src/Sel/SecretKey/EncryptedStream.hs
kozross: Does `mallocPlainForeignPtrBytes` give us a finalizer?

sel/src/Sel/SecretKey/EncryptedStream.hs
kozross: Huh, good to know.

sel/src/Sel/SecretKey/EncryptedStream.hs
```diff
@@ -0,0 +1,650 @@
...
+-- @since 0.0.1.0
+initPullStream
+  :: Multipart s
+  -- ^ Cryptographic state.
+  -> Header
+  -- ^ Header received from the peer, used to ensure the ciphertexts will be valid.
+  -> SecretKey
+  -- ^ Secret key from the peer, used to decrypt the ciphertexts.
+  -> IO (Either EncryptedStreamError ())
+  -- ^ Returns 'True' on success, 'False' if the header is invalid
```
kozross, line 519: Was this meant to return an `IO (Either EncryptedStreamError Bool)`?

sel/src/Sel/SecretKey/EncryptedStream.hs
```diff
@@ -0,0 +1,650 @@
...
+-- Applications will typically call this function in a loop,
+-- until a message with the 'Final' tag is found.
+--
+-- If the tag cannot be decoded from the payload, then it is not returned in the `StreamResult`.
+--
+-- @since 0.0.1.0
+pullFromStream
+  :: Multipart s
+  -> CipherText
+  -> Maybe Word
```
kozross, line 542: What's this argument meant to do?

kozross (changes requested):

sel/src/Sel/SecretKey/EncryptedStream.hs
Kleidukos: Actually the doc is wrong, I'm experimenting with more expressive error types.

Kleidukos:
Closing in favour of something based on a pre-existing streaming library

PR #83, opened 2023-06-04
Add `nix` support for local development

monadplus:
It features:
- Building both packages, generating the documentation and running packages tests in a reproducible environment through a single command: `nix build .#libsodium-bindings` and `nix build .#sel`, respectively.
- A _deterministic_ environment multi-platform to build both packages using Cabal: `nix develop`. This environment includes:
  - GHC 9.4.5
  - Cabal 3.10 with all required packages installed
  - The right version of HLS
  -  All formatters: cabal-fmt, fourmolu, and nixpkgs-fmt
  - All linters: hlint
  - Installed githooks that prevent committing without formatting (`*.cabal`, `*.hs`, and `*.nix`) and linting.
- Integration with [direnv](https://direnv.net/) through [nix-direnv](https://github.com/nix-community/nix-direnv).
- Backwards compatibility with `nix-build` and `nix-shell`.

It does not feature:
- Integration with CI.

This PR was motivated by my own needs. @TeofilC gently asked me to review the documentation, but I couldn't compile the packages and generate the latest documentation. Apart from that, I was lacking tooling such as cabal-fmt, fourmolu, etc. So, I decided to write a nix derivation `default.nix` to fulfill those needs (first commit). A week later, I started learning about flakes and decided to port this derivation to `flake.nix`.

This nix integration really helped me get the development environment I needed for this project without having to install all the dependencies and tooling. I thought it may also be beneficial for others, and so I talked with @TeofilC to see if this project was open to integrate with nix and they responded positively. I understand that giving support to nix may inquire a small additional maintenance cost, so up you folks whether to finally accept this MR or not. 

P.S. I have only been using flakes for a few day, so there's probably a lot of ways to improve this work, so suggestions are always welcome!

PR #84, opened 2023-06-05
fixup nix

MangoIV:
This supercedes #83 with some refactoring and simplification such as: 
- make everything compile in both packages and cabal
- refactor some of the nix
- make the package generation more reliable
- add packages and shell to checks
- reduce flake inputs
- allow for `nix flake check` with `--impure`
- reduce complexity in overlay
- added workflow to check both nix packages as well as the `devShell`
- added cachix pubkey and url in extra substitutors  

cc @monadplus


MangoIV:
This relies on IFD which is necessary to make the package generation reliable in case of the upstream version adding e.g. more dependencies than the version currently on hackage (which is very probable)

monadplus:
Thanks for fixing my mess <3

Do you think it is worth prefixing the packages by the compiler version? We could add outputs for the supported GHC versions in cabal by just parametrizing the some parts, but probably it won't be that easy to make everything compile.

MangoIV:
@Kleidukos I think this is ready for review :)

MangoIV:
@monadplus Would be happy about a review from you for at least the small parts you didn't write yourself, as well :3

monadplus:
I run all outputs on a x86-64 linux successfully. It would be useful to test on a darwin arm, but not necessary.

The changes seem reasonable. You enabled incremental compilation, improved the cabal pkg-config, and fixed the missing dyn dependency on the repl.

I see you removed some of the tools likes hlint. Is there a reason for that? I know the commit-hook run those for you, but it is still worth having them while developing. Maybe hls provides hlint? Anyway, I am ok with those changes.

Awesome work!

MangoIV:
> Is there a reason for that

There was at first, because I assumed that the only place we would need hlint is the pre-commit-hook, but you're right, we need it for hls as well, I will add it back. 

wrt the nix formatters I think it's fine to only have it in pchs, that gives it less chance for divergence

MangoIV:
at the moment I think they can't diverge because they have the same nixpkgs

MangoIV:
but theoretically they can that's why I removed most of them

MangoIV:
> `LD_LIBRARY_PATH` 

this and pkgconfigdepends

Kleidukos (approved):
let's go!

Issue #88, opened 2023-06-14
Zeroing ForeignPtrs on deallocation


dpwiz:
Given all the warnings and "[REDACTED]" instances I was surprised that there are no memory clean-up for those objects.

Perhaps there should be an `mallocForeignPtrBytes` wrapper that automatically attaches `addForeignPtrFinalizer` with `fillBytes` or `sodium_memzero()`.

OTOH, when you import your secrets from a `ByteString` or hex, it's already a game over in this department :woman_shrugging:

dpwiz:
Or just go all in on `sodium_malloc()`.

Kleidukos:
Certainly sodium_malloc() could be used, yes.
Do you feel like you'd like to tackle this? :)

tchoutri:
closed via #91

PR #91, opened 2023-07-02
Use sodium_alloc/free for SecretKey memory

dpwiz:
For #88

The changes are minimal with the further improvements possible:

- Treating SecretKey as an allocated resource instead of relying on GC.
- A proper Exception instead of `throwErrno`.
- Using `sodium_mprotect_noaccess` to guard memory at rest.
- Extract the stuff into a generic SecureMemory wrapper.


libsodium-bindings/src/LibSodium/Bindings/SecureMemory.hs
```diff
@@ -178,3 +179,11 @@ foreign import capi "sodium.h sodium_free"
...
      . Ptr a
     -> IO ()
+
+-- | Function pointer to use as 'Foreign.ForeignPtr' finalizer for sodium-allocated memory.
+--
+-- The memory region is filled with zeros before the deallocation.
+foreign import ccall "sodium.h &sodium_free"
+  finalizerSodiumFree
+    :: forall a
+     . FinalizerPtr a
```
a-02, line 189: Thanks for writing all this for us. As of right now we highly prefer `capi` over `ccall`.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -135,15 +134,38 @@ secretKeyFromByteString bytestring =
...
     then Nothing
     else unsafeDupablePerformIO $
-      BS.unsafeUseAsCStringLen bytestring $ \(outsideSecretKeyPtr, _) -> do
-        secretKeyForeignPtr <-
-          BS.mallocByteString @CChar (fromIntegral cryptoSecretboxKeyBytes)
-        Foreign.withForeignPtr secretKeyForeignPtr $ \secretKeyPtr ->
-          Foreign.copyArray
+      BS.unsafeUseAsCStringLen bytestring $ \(outsideSecretKeyPtr, _) ->
+        fmap Just $
+          newSecretKeyWith $ \secretKeyPtr ->
```
a-02, line 139: This doesn't change any behavior, right? If it's just moving things around then it's fine.

libsodium-bindings/src/LibSodium/Bindings/SecureMemory.hs
Kleidukos: Indeed, unless there is a problem with GHC, we try to stick to `capi`: https://github.com/haskell-cryptography/libsodium-bindings/blob/main/libsodium-bindings/CONTRIBUTING.md#c-ffi

dpwiz: According to [finalizer docs](https://hackage.haskell.org/package/base-4.18.0.0/docs/GHC-ForeignPtr.html#t:FinalizerPtr) this must be a `ccall`:

> Note that the foreign function must use the ccall calling convention.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
dpwiz: Unless I've missed something the old code was just unwrapping bytestring, does the same allocation routine, then copying bytes. Hence the common function.

libsodium-bindings/src/LibSodium/Bindings/SecureMemory.hs
Kleidukos: I investigated a bit and actually capi is fine too: https://gitlab.haskell.org/ghc/ghc/-/issues/23599

The thing is that the capi calling convention generates a C stub that calls to the named function (that we foreign import). The problem is with the `stdcall` and `js` conventions, where these functions get called by the RTS directly.

sel/src/Sel/SecretKey/AuthenticatedEncryption.hs
```diff
@@ -135,15 +134,38 @@ secretKeyFromByteString bytestring =
...
+-- | Prepare memory for a 'SecretKey' and use the provided action to fill it.
+--
+-- Memory is allocated with 'LibSodium.Bindings.SecureMemory.sodiumMalloc' (see the note attached there).
+-- Finalizer is run when the key is goes out of scope, but 'freeSecretKey' can be used to release early.
+--
+-- @since 0.0.1.0
+newSecretKeyWith :: (Foreign.Ptr CUChar -> IO ()) -> IO SecretKey
+newSecretKeyWith action = do
+  ptr <- sodiumMalloc cryptoSecretboxKeyBytes
+  when (ptr == Foreign.nullPtr) $ do
```
a-02, line 154: I was going to say something about using our foreign pointer equality function (`foreignPtrEq`), then I realized that since it's a null pointer, that wouldn't make sense. Looks fine.

Kleidukos (approved):

PR #94, opened 2023-07-13
Add public key authenticated encryption to Sel

### LINKED ISSUE #93: Public Key Authenticated Encryption (CryptoBox)
Write bindings for LibSodium.Bindings.CryptoBox's easy API

Kleidukos:
The test fails and it really mystifies me.
@kozross I would appreciate the help on that one. :) 

Fixes #93

### DIFF
--- a/sel/src/Sel/PublicKey/AuthenticatedEncryption.hs
+++ b/sel/src/Sel/PublicKey/AuthenticatedEncryption.hs
@@ -21,38 +21,41 @@ module Sel.PublicKey.AuthenticatedEncryption
     -- $usage
 
     -- ** Key pair generation
-    SecretKey
-  , PublicKey
-  , newKeyPair
-  , secretKeyPairFromByteStrings
-  , freeSecretKey
+    newKeyPair
+  , SecretKey
   , unsafeSecretKeyToHexByteString
-  , unsafePublicKeyToHexByteString
+  , freeSecretKey
+  , PublicKey
+  , secretKeyPairFromHexByteStrings
 
     -- ** Nonce
   , Nonce
-  , nonceFromByteString
+  , nonceFromHexByteString
   , nonceToHexByteString
 
-    -- ** Hash
-  , Hash
-  , hashFromByteString
-  , hashToHexText
-  , hashToHexByteString
-  , hashToBinary
+    -- ** Cipher text
+  , CipherText
+  , cipherTextFromHexByteString
+  , cipherTextToHexText
+  , cipherTextToHexByteString
+  , cipherTextToBinary
 
     -- ** Encryption and Decryption
   , encrypt
   , decrypt
+
+    -- ** Errors
+  , KeyPairGenerationException
   ) where
 
-import Control.Monad (void, when)
+import Control.Monad (when)
 import Data.ByteString (StrictByteString)
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Base16 as Base16
 import qualified Data.ByteString.Internal as BS
 import qualified Data.ByteString.Unsafe as BS
 import Data.Text (Text)
+import qualified Data.Text as Text
 import Data.Text.Display (Display (displayBuilder), OpaqueInstance (..), ShowInstance (..))
 import qualified Data.Text.Lazy.Builder as Builder
 import Data.Word (Word8)
@@ -66,7 +69,7 @@ import System.IO.Unsafe (unsafeDupablePerformIO)
 import Control.Exception
 import LibSodium.Bindings.CryptoBox
 import LibSodium.Bindings.Random (randombytesBuf)
-import LibSodium.Bindings.SecureMemory
+import LibSodium.Bindings.SecureMemory (finalizerSodiumFree, sodiumFree, sodiumMalloc)
 import Sel.Internal
 
 -- $introduction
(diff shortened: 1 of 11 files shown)


sel/src/Sel/PublicKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,494 @@
...
+data KeyPairGenerationException = KeyPairGenerationException
+  deriving stock (Eq, Ord, Show)
+  deriving anyclass (Exception)
+
+-- | Generate a new random secret key.
+--
+-- @since 0.0.1.0
+newKeyPair :: IO (SecretKey, PublicKey)
+newKeyPair = newKeyPairWith $ \secretKeyPtr publicKeyPtr -> do
+  result <- cryptoBoxKeyPair secretKeyPtr publicKeyPtr
```
int-e, line 165: These arguments are in the wrong order.

sel/src/Sel/PublicKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,494 @@
...
+       -> Ptr CUChar
+       -- \^ Public Key pointer
+       -> IO ()
+     )
+  -> IO (SecretKey, PublicKey)
+newKeyPairWith action = do
+  secretKeyPtr <- sodiumMalloc cryptoBoxSecretKeyBytes
+  publicKeyPtr <- sodiumMalloc cryptoBoxPublicKeyBytes
+  when (secretKeyPtr == Foreign.nullPtr || publicKeyPtr == Foreign.nullPtr) $ do
+    Foreign.throwErrno "sodium_malloc"
```
int-e, line 216: If the second `malloc` fails, the first allocation will leak.

sel/src/Sel/PublicKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,494 @@
...
+-- Memory is allocated with 'LibSodium.Bindings.SecureMemory.sodiumMalloc' (see the note attached there).
+-- Finalizer is run when the key is goes out of scope, but 'freeSecretKey' can be used to release early.
+--
+-- @since 0.0.1.0
+newKeyPairWith
+  :: ( Ptr CUChar
+       -- \^ Secret Key pointer
+       -> Ptr CUChar
+       -- \^ Public Key pointer
+       -> IO ()
```
int-e, line 209: Since libsodium always puts the public key first, it's probably best to follow that convention for the callback as well.

sel/src/Sel/PublicKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,494 @@
...
+nonceToHexByteString (Nonce nonceForeignPtr) =
+  Base16.encodeBase16' $
+    BS.fromForeignPtr0
+      (Foreign.castForeignPtr @CUChar @Word8 nonceForeignPtr)
+      (fromIntegral @CSize @Int cryptoBoxNonceBytes)
+
+-- | A ciphertext consisting of an encrypted message and an authentication tag.
+--
+-- @since 0.0.1.0
+data Hash = Hash
```
int-e, line 336: It feels kind of strange to me to call a ciphertext a hash.

Apart from that, I think it might be a good idea to include the nonce with the message so that the association between the nonce and the message is not easily lost?

int-e (commented):
I found the bug in `newKeyPair`, see review comments for details.

sel/src/Sel/PublicKey/AuthenticatedEncryption.hs
```diff
@@ -0,0 +1,494 @@
...
+                hashPtr
+                (messageLength + fromIntegral cryptoBoxMACBytes)
+                noncePtr
+                publicKeyPtr
+                secretKeyPtr
+            case result of
+              (-1) -> pure Nothing
+              _ -> do
+                bsPtr <- Foreign.mallocBytes (fromIntegral messageLength)
+                memcpy bsPtr (Foreign.castPtr messagePtr) (fromIntegral messageLength)
```
int-e, line 491: Is this copy needed? AFAICS `messagePtr` can be used directly if one changes `CChar` to `CUChar` below.

Kleidukos:
@int-e you rock, thanks!

sel/src/Sel/PublicKey/AuthenticatedEncryption.hs
Kleidukos: Thanks!

sel/src/Sel/PublicKey/AuthenticatedEncryption.hs
Kleidukos: well-spotted! Would this be better?

```haskell
when (secretKeyPtr == Foreign.nullPtr || publicKeyPtr == Foreign.nullPtr) $ do
  sodiumFree secretKeyPtr
  sodiumFree publicKeyPtr
  Foreign.throwErrno "sodium_malloc"
```

sel/src/Sel/PublicKey/AuthenticatedEncryption.hs
Kleidukos: Yep, renaming it.

sel/src/Sel/PublicKey/AuthenticatedEncryption.hs
Kleidukos: Taking into account the dangerous nature of `unsafePackMallocCStringLen` I thought it would be safer. But also I have much to learn in terms of determining when boundaries are violated.

PR #97, opened 2023-08-04
Add salt conversion functions

RaoulHC:
Salts often need to be stored alongside passwords so the opaque representation needs a way of being converted to something we can deal with. Constructing salts checks the 'StrictByteString' is the right length.

While reading `ByteString` documentation I also realised that `genSalt` could be much more neatly implemented using `create`, and I added a bunch of tests for using the `hashByteStringWithParams`.


Kleidukos (approved):

Kleidukos:
LGTM, thanks a lot @RaoulHC !

PR #99, opened 2023-08-04
Enforce hexadecimal encoding of 'PasswordHash'

### LINKED ISSUE #98: `passwordHashToText` is broken
`passwordHashToText` doesn't work because there's no guarantee that the hash generated is valid UTF-8.

It should either be removed or encoded in Base16 as a lot of the hashes in the rest of `sel` seem to do, though in the latter case it should probably be renamed `passwordHashToHexText`.

Kleidukos:
fixes #98

### DIFF
--- a/sel/src/Sel/Hashing/Password.hs
+++ b/sel/src/Sel/Hashing/Password.hs
@@ -23,14 +23,19 @@ module Sel.Hashing.Password
   , hashByteStringWithParams
 
     -- ** Conversion to textual formats
-  , passwordHashToByteString
-  , passwordHashToText
+  , passwordHashToBinary
+  , passwordHashToHexText
+  , passwordHashToHexByteString
 
     -- * Salt
   , Salt
   , genSalt
-  , saltToByteString
-  , byteStringToSalt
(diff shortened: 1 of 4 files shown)


Kleidukos:
@RaoulHC does this look good?

RaoulHC:
I don't think you want to get rid of the original version of this, encoding doesn't matter for `ByteString`s and leaving a non hex version allows users to encode it how they want. I see `Sel.Hashing.SHA2.SHA512` has both `hashToBinary` and `hashToHexByteString` so perhaps something along those lines makes the most sense.

I also notice in both `SHA2` modules the `HexText` functions aren't encoding the binary into base16 and appear to be faulty, so probably makes sense to fix them in this PR too.

Kleidukos:
@RaoulHC Great, thank you!

Kleidukos:
@RaoulHC Hopefully this should be good now

Kleidukos:
@RaoulHC Thanks for the report and the review btw, it's highly appreciated. Don't hesitate to point out other inconsistencies if you see them (or ping on Matrix)

PR #101, opened 2023-08-08
Add ShortHash to Sel

### LINKED ISSUE #90: ShortHash for sel
Sel should have a module for https://github.com/haskell-cryptography/libsodium-bindings/blob/main/libsodium-bindings/src/LibSodium/Bindings/ShortHashing.hs

Kleidukos:
Fix #90

* [ ] "Usage" section

Screenshot 2023-08-09 at 01-03-20 Sel Hashing Short


sel/src/Sel/Hashing/Short.hs
```diff
@@ -0,0 +1,297 @@
...
+-- * Integrity checking in interactive protocols
+
+-- | A 128-bit hash of a short input, of size 'cryptoShortHashSipHashX24Bytes'
+--
+-- @since 0.0.1.0
+newtype ShortHash = ShortHash (ForeignPtr CUChar)
+  deriving stock
+    ( Show
+      -- ^ @since 0.0.1.0
+    )
```
kozross, line 87: This'll just spew an address. Is that what we want for the `Show` instance?

sel/src/Sel/Hashing/Short.hs
```diff
@@ -0,0 +1,297 @@
...
+-- | A random key used for hashing, of size 'cryptoShortHashSipHashX24KeyBytes'.
+--
+-- The same message hashed with the same key will always produce the same output.
+--
+-- @since 0.0.1.0
+newtype ShortHashKey = ShortHashKey (ForeignPtr CUChar)
+  deriving stock
+    ( Show
+      -- ^ @since 0.0.1.0
+    )
```
kozross, line 184: Same here.

kozross (changes requested):

sel/src/Sel/Hashing/Short.hs
Kleidukos: Yes I know, I thought for a bit that it would be relevant to have `Show` instance display the internal of the structure, and Display show a more user-facing representation (hexadecimal), but perhaps it's not really a problem I should concern myself with and basically have Show == Display?

Kleidukos: Alrighty

kozross (approved):

PR #104, opened 2023-08-28
Add Eq, Ord, Show instances for PasswordHash and Salt.

a-02:


sel/src/Sel/Hashing/Password.hs
```diff
@@ -181,6 +196,7 @@ passwordHashToHexText = Base16.encodeBase16 . passwordHashToBinary
 --
 -- @since 0.0.1.0
 newtype Salt = Salt StrictByteString
+  deriving (Eq, Ord, Show)
```
kozross, line 199: It would be better for this to use explicit strategies.

kozross (changes requested):

sel/src/Sel/Hashing/Password.hs
```diff
@@ -74,6 +76,19 @@ import LibSodium.Bindings.Random
...
+      foreignPtrEq ph1 ph2 cryptoPWHashStrBytes
+
+instance Ord PasswordHash where
+  (PasswordHash ph1) `compare` (PasswordHash ph2) =
+    unsafeDupablePerformIO $
+      foreignPtrOrd ph1 ph2 cryptoPWHashStrBytes
+
+instance Show PasswordHash where
+  show (PasswordHash fptr) = foreignPtrShow fptr cryptoPWHashStrBytes
+
```
Kleidukos, line 91: If you could add:

```
-- |
--
-- @since 0.0.1.0
```

to the instances that would be perfect

Kleidukos (changes requested):

sel/src/Sel/Hashing/Password.hs
```diff
@@ -182,6 +200,18 @@ passwordHashToHexText = Base16.encodeBase16 . passwordHashToBinary
...
+instance Eq Salt where
+  (Salt s1) == (Salt s2) = s1 == s2
+
+-- | @since 0.0.1.0
+instance Ord Salt where
+  (Salt s1) `compare` (Salt s2) = s1 `compare` s2
+
+-- | @since 0.0.1.0
+instance Show Salt where
+  show (Salt s1) = show s1
```
Kleidukos, line 213: I believe there's been a misunderstanding, "explicit strategies" refers to this: https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/deriving_strategies.html

a-02: Ah. Sorry about that.

Kleidukos: no worries. :)

a-02:
Everything's in, just can't get fourmolu to play nice locally.

Kleidukos:
@a-02 what's the exact problem? Are you using fourmolu-0.12.0.0?

a-02:
That might be it. I'm trying to lint things on my end but it keeps coming back as "all good".

Kleidukos:
@a-02 I've tried something, let's see if it works

Kleidukos (approved):

Kleidukos:
@kozross I think we're good for this PR :+1:

sel/sel.cabal
```diff
@@ -33,7 +33,12 @@ common common
     -fprint-potential-instances -Wno-unticked-promoted-constructors
     -Werror=unused-imports
 
-  default-language: Haskell2010
+  default-language:   Haskell2010
+  default-extensions:
+    DerivingStrategies
+    GeneralizedNewtypeDeriving
+    ScopedTypeVariables
+    TypeApplications
```
kozross, line 41: Could we not do this please? Having extensions manually specified per-file is much clearer to a reader.

kozross (changes requested):

sel/src/Sel/Hashing/Password.hs
```diff
@@ -228,6 +240,7 @@ asciiByteStringToPasswordHash textualHash =
 --
 -- @since 0.0.1.0
 newtype Salt = Salt StrictByteString
+  deriving newtype (Eq, Ord, Show)
```
kozross, line 243: Each of these needs `@since`s too.

kozross (changes requested):

kozross (approved):

PR #110, opened 2023-09-03
Add more conversion functions for PasswordHash

Kleidukos:
The following functions are added:

* `asciiTextToPasswordHash`
* `asciiByteStringToPasswordHash`

The idea is that there is this password ASCII format that looks like this:

`$argon2id$v=19$m=262144,t=3,p=1$fpPdXj9mK7J4m…`

Password hashes are created by `hashByteString` in this format.

Keys that are derived with `hashByteStringWithParams` are not stored like this. It's a binary format that needs base16 encoding. I think I'm not too comfortable with this, we should probably implement the ASCII format for the output of this function.


Kleidukos:
@kozross any remark?

kozross (approved):

PR #114, opened 2023-09-22
Use the ASCII encoding of PasswordHash in its Show instance

Kleidukos:
This reduces surprise when people try it out in ghci


a-02 (approved):
Yeah, this is a good idea. Do the other uses of `foreignPtrShow` need to display the ASCII contents?

Kleidukos:
> Yeah, this is a good idea. Do the other uses of `foreignPtrShow` need to display the ASCII contents?

Not necessarily. Password hashes are a special case.

Kleidukos:
@Mergifyio refresh

PR #117, opened 2023-10-08
Support newer GHC versions

Kleidukos:


flake.nix
```diff
@@ -73,6 +75,7 @@
               libsodium-bindings = commonOverrides (hself.callCabal2nix "libsodium-bindings" (gitignoreSource ./libsodium-bindings) { });
               sel = commonOverrides (hself.callCabal2nix "sel" (gitignoreSource ./sel) { });
               text-display = markUnbroken hsuper.text-display;
+              base16 = inputs.base16;
```
MangoIV, line 78: base16 should be a derivation that contains a haskell package, more specifically it should not be a store path to the source of base qy as it is currently. 
```suggestion
              base16 = hself.callCabal2nix "base16" (gitignoreSource inputs.base16);
```

Kleidukos: I see

flake.nix
```diff
@@ -73,6 +75,7 @@
               libsodium-bindings = commonOverrides (hself.callCabal2nix "libsodium-bindings" (gitignoreSource ./libsodium-bindings) { });
               sel = commonOverrides (hself.callCabal2nix "sel" (gitignoreSource ./sel) { });
               text-display = markUnbroken hsuper.text-display;
+              base16 = hself.callCabal2nix "base16" (gitignoreSource inputs.base16);
```
MangoIV, line 78:
```suggestion
              base16 = hself.callCabal2nix "base16" (gitignoreSource inputs.base16) { };
```

PR #123, opened 2023-11-05
Remove manual freeing functions from Sel

Kleidukos:
This PR removes low-level deallocation functions from the high-level API.


Kleidukos:
@clementinakassiopeia what do you think?

ClementinaKassiopeia:
Looks good! Though I would also hide [newSecretKeyWith](https://hackage.haskell.org/package/sel-0.0.1.0/candidate/docs/Sel-SecretKey-Cipher.html#v:newSecretKeyWith) from Sel.SecretKey.Cipher.

Issue #131, opened 2023-12-07
Unable to re-use nix-based build


blackheaven:
I'm trying to add [nix build support on a project which depends on `sel`](https://github.com/haskell-cryptography/one-time-password/pull/6), but I fail to make *GHC* detect `sel`:

```
@nix { "action": "setPhase", "phase": "setupCompilerEnvironmentPhase" }
setupCompilerEnvironmentPhase
Build with /nix/store/psds2pz1qhlr4z8qcahqii6kq1xsawb8-ghc-9.4.8.
@nix { "action": "setPhase", "phase": "unpackPhase" }
unpacking sources
unpacking source archive /nix/store/sbin4qzj03wn252n0qibwn6jj9p1xnrc-bvh2s0sx28rfp0gzysj4p5fwzxzz0djx-source
source root is bvh2s0sx28rfp0gzysj4p5fwzxzz0djx-source
@nix { "action": "setPhase", "phase": "patchPhase" }
patching sources
@nix { "action": "setPhase", "phase": "compileBuildDriverPhase" }
compileBuildDriverPhase
setupCompileFlags: -package-db=/build/tmp.2qEJsW8rMl/setup-package.conf.d -j16 +RTS -A64M -RTS -threaded -rtsopts
[1 of 2] Compiling Main             ( /nix/store/4mdp8nhyfddh7bllbi7xszz7k9955n79-Setup.hs, /build/tmp.2qEJsW8rMl/Main.o )
[2 of 2] Linking Setup
@nix { "action": "setPhase", "phase": "updateAutotoolsGnuConfigScriptsPhase" }
updateAutotoolsGnuConfigScriptsPhase
@nix { "action": "setPhase", "phase": "configurePhase" }
configuring
[...]

PR #132, opened 2023-12-08
fix: use current GHC in flake (#131)

blackheaven:
Users would then be able to change `inputs.nixpkgs` and select the right version.

Alternatively I can define one attribute per version.

Let me know what you think.


Kleidukos:
I don't really care, as long as we can test all the versions declared as `tested-with` it's alright for me. :)

Kleidukos:
Fantastic, thank you @blackheaven!

PR #168, opened 2025-01-03
some cleanups

MangoIV:
- remove some files that were left over when the nix support was removed
- allow more tasty versions (sel is broken on nixpkgs because tasty hasn't moved along yet)


Kleidukos (approved):
Merci !

PR #169, opened 2025-01-03
feat(Internal): add constant-time Eq, use Scoped for internals

jhenahan:
Constant-time comparison helps avoid timing attacks on secret cryptographic material.

Also adds functions for interacting with `sodium_malloc`'d pointers.

Stack
- #169 <- you are here
- https://github.com/jhenahan/libsodium-bindings/pull/2
- https://github.com/jhenahan/libsodium-bindings/pull/3

### DIFF
--- a/sel/src/Sel/Hashing/Password.hs
+++ b/sel/src/Sel/Hashing/Password.hs
@@ -94,20 +94,18 @@ instance Display PasswordHash where
 -- | @since 0.0.1.0
 instance Eq PasswordHash where
   (PasswordHash ph1) == (PasswordHash ph2) =
-    unsafeDupablePerformIO $
-      foreignPtrEq
-        (Foreign.castForeignPtr @CChar @CUChar ph1)
-        (Foreign.castForeignPtr @CChar @CUChar ph2)
-        cryptoPWHashStrBytes
+    foreignPtrEq
+      (Foreign.castForeignPtr @CChar @CUChar ph1)
+      (Foreign.castForeignPtr @CChar @CUChar ph2)
+      cryptoPWHashStrBytes
 
 -- | @since 0.0.1.0
 instance Ord PasswordHash where
   (PasswordHash ph1) `compare` (PasswordHash ph2) =
-    unsafeDupablePerformIO $
-      foreignPtrOrd
-        (Foreign.castForeignPtr @CChar @CUChar ph1)
-        (Foreign.castForeignPtr @CChar @CUChar ph2)
-        cryptoPWHashStrBytes
+    foreignPtrOrd
+      (Foreign.castForeignPtr @CChar @CUChar ph1)
+      (Foreign.castForeignPtr @CChar @CUChar ph2)
+      cryptoPWHashStrBytes
 
 -- | @since 0.0.1.0
 instance Show PasswordHash where
--- a/sel/src/Sel/Hashing/Password.hs
+++ b/sel/src/Sel/Hashing/Password.hs
@@ -95,13 +95,19 @@ instance Display PasswordHash where
 instance Eq PasswordHash where
   (PasswordHash ph1) == (PasswordHash ph2) =
     unsafeDupablePerformIO $
-      foreignPtrEq ph1 ph2 cryptoPWHashStrBytes
+      foreignPtrEq
+        (Foreign.castForeignPtr @CChar @CUChar ph1)
+        (Foreign.castForeignPtr @CChar @CUChar ph2)
+        cryptoPWHashStrBytes
 
 -- | @since 0.0.1.0
 instance Ord PasswordHash where
   (PasswordHash ph1) `compare` (PasswordHash ph2) =
     unsafeDupablePerformIO $
-      foreignPtrOrd ph1 ph2 cryptoPWHashStrBytes
+      foreignPtrOrd
+        (Foreign.castForeignPtr @CChar @CUChar ph1)
+        (Foreign.castForeignPtr @CChar @CUChar ph2)
+        cryptoPWHashStrBytes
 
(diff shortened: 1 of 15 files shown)


sel/src/Sel/Hashing/Password.hs
```diff
@@ -95,13 +95,13 @@ instance Display PasswordHash where
 instance Eq PasswordHash where
   (PasswordHash ph1) == (PasswordHash ph2) =
     unsafeDupablePerformIO $
-      foreignPtrEq ph1 ph2 cryptoPWHashStrBytes
+      foreignPtrEq (Foreign.castForeignPtr ph1) (Foreign.castForeignPtr ph2) cryptoPWHashStrBytes
```
Kleidukos, line 98: since casting functions like `fromIntegral`, `castForeignPtr` `castPt` are the Devil, I'd like it if you could insert type applications in order to make sure that there is no inference happening and that they are converting to and from the types we expect.

sel/src/Sel/Internal.hs
```diff
@@ -1,53 +1,103 @@
...
+    ByteString.fromForeignPtr cstring 0 (fromIntegral size)
+
+-- | Copy a byte array as key material.
+--
+-- The size of the array is not checked. The input may be truncated if
+-- it is too long, or an unchecked exception may be thrown if it is
+-- too short.
+--
+-- @since 0.0.3.0
+unsafeCopyToSodiumPointer :: CSize -> StrictByteString -> IO (ForeignPtr CUChar)
```
jhenahan, line 78: This isn't used in this PR, but is used in the next PR in the stack: https://github.com/jhenahan/libsodium-bindings/pull/2/files#diff-6d90c4955b4c0bcc7d190b7109069933f66133e3a46f1fa0a64e6308f33f4707R499

sel/src/Sel/Internal.hs
```diff
@@ -1,53 +1,103 @@
...
-      do
-        result <- memcmp p q size
-        return $ 0 == result
+-- | Lexicographically compare the contents of two byte arrays.
+--
+-- ⚠️ Such comparisons are vulnerable to timing attacks, and should be
+-- avoided for secret data.
+--
+-- @since 0.0.1.0
+foreignPtrOrd :: ForeignPtr CUChar -> ForeignPtr CUChar -> CSize -> IO Ordering
```
Kleidukos, line 45: I'd like your opinion on this, but considering that we only manipulate data we manage, couldn't this function be of return type `Ordering` instead of `IO Ordering`?

sel/src/Sel/Internal.hs
```diff
@@ -1,53 +1,103 @@
...
-            | result == 0 -> EQ
-            | result < 0 -> LT
-            | otherwise -> GT
+-- | Compare two byte arrays for lexicographic equality.
+--
+-- ⚠️ Such comparisons are vulnerable to timing attacks, and should be
+-- avoided for secret data.
+--
+-- @since 0.0.1.0
+foreignPtrEq :: ForeignPtr CUChar -> ForeignPtr CUChar -> CSize -> IO Bool
```
Kleidukos, line 59: Same question as for `foreignPtrOrd`

sel/src/Sel/Internal.hs
```diff
@@ -1,53 +1,103 @@
...
+-- and render the hexadecimal-encoded bytes as a 'String'.
+--
+-- @since 0.0.1.0
 foreignPtrShow :: ForeignPtr a -> CSize -> String
-foreignPtrShow fptr size =
-  BS.unpackChars . Base16.extractBase16 . Base16.encodeBase16' $
-    BS.fromForeignPtr (Foreign.castForeignPtr fptr) 0 (fromIntegral @CSize @Int size)
+foreignPtrShow (Foreign.castForeignPtr -> cstring) size =
+  ByteString.unpackChars . Base16.extractBase16 . Base16.encodeBase16' $
+    ByteString.fromForeignPtr cstring 0 (fromIntegral size)
```
Kleidukos, line 69: Please put back the type applications.

sel/src/Sel/Internal.hs
```diff
@@ -1,53 +1,103 @@
...
+
+-- | Copy a byte array as key material.
```
Kleidukos, line 71: What does "as key material" mean here? Are there invariants not to break, is this to encourage contributors to use this function for keys only?

sel/src/Sel/Scrypt.hs
Kleidukos, line 1: Casting functions that take polymorphic arguments should have type applications.

Kleidukos (changes requested):

sel/src/Sel/Internal.hs
jhenahan: Certainly could be. In the existing code the practice seems to be to push the `unsafeDupablePerformIO` closer to the consumer, but I'm not at all opposed to just doing it here.

sel/src/Sel/Internal.hs
jhenahan: Partially a hedge, partially an artifact of when this was doing more work via `coerce` in the previous version of this work and was explicitly producing keys. I'll reword to refer to sodium pointers rather than explicitly key material, though the current _usecase_ is definitely just keys. I suppose there's no reason one _has_ to use it for key material, though.

sel/src/Sel/Internal.hs
Kleidukos: Yeah let's not be too rigid, you can put the `unsafeDupablePerformIO` here. :)

Kleidukos (approved):

Kleidukos:
Cheers

PR #170, opened 2025-01-03
feat(Signature): refactor and expand public-key signature API

jhenahan:
Highlights:

- Models input validation errors (pointer length and hex encoding)

- Serde and rendering functions for `PublicKey`, `SecretKey`, and `SignedMessage`

- Internals under `Scoped`

- Uses constant-time `Eq` and `Ord` for `SecretKey`

- For consistency, use `Foreign.copyBytes` everywhere that `memcpy` was once used

### DIFF
--- a/sel/src/Sel/PublicKey/Internal/Signature.hs
+++ b/sel/src/Sel/PublicKey/Internal/Signature.hs
@@ -1,0 +1,451 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveTraversable #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE ViewPatterns #-}
+
+module Sel.PublicKey.Internal.Signature
+  ( -- ** Public Keys
+    PublicKey
+  , decodePublicKeyHexByteString
+  , encodePublicKeyHexByteString
+
+    -- ** Secret Keys
+  , SecretKey
+  , decodeSecretKeyHexByteString
+  , unsafeEncodeSecretKeyHexByteString
+  , publicKey
+
+    -- ** Key Pairs
+  , KeyPair (..)
+  , keyPair
+
+    -- ** Signed Messages
+  , SignedMessage
+  , sign
+  , open
+  , SignatureVerification (..)
+  , extractUnverifiedMessage
+  , extractSignature
+  , buildSignedMessage
+
+    -- ** Exceptions
+  , PublicKeyExtractionException (..)
+  )
+where
+
+import Control.Exception (Exception, throw)
+import Control.Monad (unless)
+import Control.Monad.Trans.Class (lift)
+import Data.Base16.Types qualified as Base16
+import Data.ByteString (StrictByteString)
+import Data.ByteString.Base16 qualified as Base16
+import Data.ByteString.Internal qualified as ByteString
+import Data.ByteString.Unsafe qualified as ByteString
+import Data.Coerce (coerce)
+import Data.Text.Display (Display, OpaqueInstance (..), ShowInstance (..))
+import Data.Traversable (for)
+import Foreign (ForeignPtr)
+import Foreign qualified
+import Foreign.C (CChar, CSize, CUChar, CULLong)
+import LibSodium.Bindings.CryptoSign
+  ( cryptoSignBytes
+  , cryptoSignDetached
+  , cryptoSignED25519SkToPk
+  , cryptoSignKeyPair
+  , cryptoSignPublicKeyBytes
+  , cryptoSignSecretKeyBytes
+  , cryptoSignVerifyDetached
+  )
+import Sel.Internal
+  ( foreignPtrEq
+  , foreignPtrEqConstantTime
+  , foreignPtrOrd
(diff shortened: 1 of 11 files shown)


sel/src/Sel/PublicKey/Internal/Signature.hs
```diff
@@ -0,0 +1,499 @@
...
+  show = ByteString.unpackChars . encodeSecretKeyHexByteString
+
+-- | By lexicographical comparison of pointer contents.
+--
+-- ⚠️ Vulnerable to timing attacks!
+--
+-- @since 0.0.3.0
+instance Ord UnsafeSecretKey where
+  a `compare` b =
+    foreignPtrOrd (coerce a) (coerce b) cryptoSignSecretKeyBytes
```
Kleidukos, line 228: This feels a bit over-engineered. `encodeSecretKeyHexByteString` ought to be a much simpler interface, no? I don't understand if there are invariants that are preserved by this type and its instances.

sel/src/Sel/PublicKey/Internal/Signature.hs
```diff
@@ -0,0 +1,499 @@
...
+
+-- | By lexicographical comparison of key pointer contents.
+--
+-- ⚠️ Vulnerable to timing attacks!
+--
+-- @since 0.0.3.0
+instance Ord KeyPair where
+  compare kp1 kp2 =
+    compare kp1.public kp2.public
+      <> comparing UnsafeSecretKey kp1.secret kp2.secret
```
Kleidukos, line 261: Since it's a default instance, people ought to be able to use it (even indirectly) without having to think twice about timing attacks. Is it possible to make use of the time-safe functions?

Kleidukos (changes requested):

sel/src/Sel/PublicKey/Internal/Signature.hs
jhenahan: It's intended as a guardrail to prevent you from doing things like serializing secret keys and using `Ord` in a way that's vulnerable to timing attacks without some kind of signal to readers that you're doing something that demands some extra attention.

sel/src/Sel/PublicKey/Internal/Signature.hs
jhenahan: https://libsodium.gitbook.io/doc/helpers#constant-time-test-for-equality is what we have in the existing bindings for constant-time comparisons, but that only gives us equal/not-equal.

It looks like I _might_ be able to use https://libsodium.gitbook.io/doc/helpers#comparing-large-numbers, but there's no binding for that one in `libsodium-bindings` at the moment. Happy to add the binding and use that instead of the `UnsafeSecretKey` if that'd work for you.

sel/src/Sel/PublicKey/Internal/Signature.hs
Kleidukos: We do already have a (albeit low-tech) mechanism for signalling unsafe operations: 
image

sel/src/Sel/PublicKey/Internal/Signature.hs
Kleidukos: I think that it would work for me yes. :)

sel/src/Sel/PublicKey/Internal/Signature.hs
jhenahan: I tend to prefer sticking such things behind a gate that's very slightly higher than that since `unsafe` is a very overloaded word. To my mind, it implies there's some `safeSecretKeyToHexByteString` waiting to be discovered, and there just isn't. Disjoint concepts <=> disjoint types. I'm not married to it, though, and with constant-time `Ord` coming quite soon I could be convinced the power-to-weight isn't there.

jhenahan: That option _is_ available in the present API as `unsafeSecretKeyHexByteString`, which just wraps whatever `SecretKey` in the newtype first.

jhenahan: Pushed commits to strip that all out. Will update the API golden test once #171 lands and I rebase this PR on top of it.

sel/src/Sel/PublicKey/Signature.hs
```diff
@@ -1,240 +1,218 @@
...
+-- extreme care. Leaking a secret key allows anyone to impersonate the
+-- creator of that key and sign messages with their identity.
+--
+-- If a secret key is compromised, all messages signed by that key
+-- should be considered compromised.
+--
+-- Secret keys are compared for equality using the constant-time
+-- 'LibSodium.Bindings.Comparison.sodiumMemcmp' to avoid timing attacks.
+--
+-- __NB:__ Prefer being explicit with t'UnsafeSecretKey' to signal your
```
jhenahan, line 101: Leftover comment. Will remove.

sel/src/Sel/PublicKey/Internal/Signature.hs
```diff
@@ -0,0 +1,470 @@
...
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE NoFieldSelectors #-}
+
+module Sel.PublicKey.Internal.Signature
+  ( -- ** Public Keys
+    type PublicKey
```
Kleidukos, line 18: What's the need for explicit namespaces here? Is there an ambiguity problem?

jhenahan: None, just being explicit. Removed.

PR #173, opened 2025-01-06
feat(Signature): add serde functions, hide internals

jhenahan:
Public API
----------

- Adds `KeyMaterialDecodeError` to signal issues during deserialization

- Adds serde functions for public and secret keys

- Adds a function to extract the public key from a secret key, with `PublicKeyExtraction` on failure

- Adds `Show` and `Display` instances for `PublicKey`, `SecretKey`, and `SignedMessage`

- Re-exports `PublicKey`, `SecretKey`, and `SignedMessage` from the internal module, but no other visible change

- Expands documentation

Internals
---------

- Moves signature internals to an internal module using the new `Scoped` machinery; this is only _really_ required because we need to get at the raw pointers to do anything interesting, but the guts don't need to be on display for the public API; if this is disagreeable, it can be moved wholesale into the public module

- Adds a (presently only internal) mechanism for tracking verification of signatures; this is intended to become public as part of an [RFC][RFC] on the API, but remains internal (and may even be removed) pending feedback

- Adds `KeyPair`, just a name for the existing `(PublicKey, SecretKey)` representation; this is also intended to become public as part of the [RFC][RFC]

[RFC]: https://github.com/haskell-cryptography/libsodium-bindings/issues/174

### DIFF
--- a/sel/src/Sel/KeyMaterialDecodeError.hs
+++ b/sel/src/Sel/KeyMaterialDecodeError.hs
@@ -1,0 +1,91 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE ViewPatterns #-}
+
+-- |
+-- Module      : Sel.KeyMaterialDecodeError
+-- Description : Key material utilities
+-- Copyright   : (c) Jack Henahan, 2024
+-- License     : BSD-3-Clause
+-- Maintainer  : The Haskell Cryptography Group
+-- Portability : GHC only
+module Sel.KeyMaterialDecodeError
+  ( -- * Key material utilities
+    KeyMaterialDecodeError (..)
+  , RequiredLength (..)
+  , InputLength (..)
+  , validKeyMaterial
+  )
+where
+
+import Control.Exception (Exception)
+import Data.Bifunctor (first)
+import Data.ByteString (StrictByteString)
+import Data.ByteString qualified as ByteString
+import Data.ByteString.Base16 qualified as Base16
+import Data.Coerce (coerce)
+import Data.Text (Text)
+import Data.Text.Display (Display, ShowInstance (..))
+import Foreign.C (CSize (..))
+
+-- | Errors arising from decoding key material from bytes.
+--
+-- @since 0.0.3.0
+data KeyMaterialDecodeError
+  = -- | Input length does not match the length required for the target pointer.
+    --
+    -- @since 0.0.3.0
+    ByteLengthMismatch RequiredLength InputLength
+  | -- | Input bytes did not decode to hexadecimal.
+    --
+    -- @since 0.0.3.0
+    DecodingFailure Text
+  deriving stock
+    ( Show
+      -- ^ @since 0.0.3.0
+    , Eq
+      -- ^ @since 0.0.3.0
+    )
+  deriving
+    ( Display
+      -- ^ @since 0.0.3.0
+    )
+    via (ShowInstance KeyMaterialDecodeError)
+  deriving anyclass
+    ( Exception
+      -- ^ @since 0.0.3.0
+    )
+
+-- | The length of the target pointer for some key material.
+--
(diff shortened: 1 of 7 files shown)


sel/src/Sel/PublicKey/Internal/Signature.hs
```diff
@@ -0,0 +1,457 @@
...
+--
+-- @since 0.0.3.0
+decodePublicKeyHexByteString :: StrictByteString -> Either KeyMaterialDecodeError PublicKey
+decodePublicKeyHexByteString bytes = PublicKey <$> copyHexKey cryptoSignPublicKeyBytes bytes
+
+-- | Encode a t'PublicKey' to a hexadecimal encoded 'StrictByteString'.
+--
+-- @since 0.0.3.0
+encodePublicKeyHexByteString :: PublicKey -> StrictByteString
+encodePublicKeyHexByteString (PublicKey publicKeyPtr) =
```
jhenahan, line 125: Looking a bit more into the bindings, it seems like we can do less work and get better security properties using the constant-time `sodium_bin2hex`. https://github.com/haskell-cryptography/libsodium-bindings/blob/5a854fff18d30402861a3fe4b56a81f1a14789a0/libsodium-bindings/src/LibSodium/Bindings/Utils.hs#L59-L71

If the `sodiumHex2Bin` binding is also revivable since the [GHC issue](https://gitlab.haskell.org/ghc/ghc/-/issues/15531) mentioned in #8 (since fixed and the fix backported), we may be able to save some implementation effort on the decoding end, as well.

jhenahan: Tried it out:

```hs
encodePublicKeyHexByteString :: PublicKey -> StrictByteString
encodePublicKeyHexByteString (PublicKey publicKeyPtr) =
  Base16.extractBase16 . Base16.encodeBase16' $
    ByteString.fromForeignPtr0
      (Foreign.castForeignPtr publicKeyPtr)
      (fromIntegral cryptoSignPublicKeyBytes)

encodePublicKeyHexByteString' :: PublicKey -> StrictByteString
encodePublicKeyHexByteString' (PublicKey publicKeyPtr) = unsafeDupablePerformIO . use $ do
  let hexLength = cryptoSignPublicKeyBytes * 2 + 1
  hexPtr <- mallocForeignPtrBytes (fromIntegral hexLength) >>= foreignPtr
  pkPtr <- foreignPtr publicKeyPtr
  lift $ ByteString.packCString =<< sodiumBin2Hex hexPtr hexLength pkPtr cryptoSignPublicKeyBytes
```

```
ghci> kp <- generateKeyPair
ghci> encodePublicKeyHexByteString (fst kp)
"03cf60081dfce4bfd6c4d7db9944c5dc77512b04c657bccbe24ae528ea5620d2"
ghci> encodePublicKeyHexByteString' (fst kp)
"03cf60081dfce4bfd6c4d7db9944c5dc77512b04c657bccbe24ae528ea5620d2"
```

Kleidukos: Perfect let's go with it then. :)

sel/src/Sel/KeyMaterialDecodeError.hs
```diff
@@ -0,0 +1,91 @@
...
+-- | Attempt to decode a hexadecimal-encoded 'StrictByteString' with an expected length.
+--
+-- @since 0.0.3.0
+validKeyMaterial :: CSize -> StrictByteString -> Either KeyMaterialDecodeError StrictByteString
+validKeyMaterial (fromIntegral -> requiredLength) bytes = do
+  decoded@(ByteString.length -> inputLength) <-
+    first DecodingFailure (Base16.decodeBase16Untyped bytes)
+  if requiredLength == inputLength
+    then Right decoded
+    else Left $ ByteLengthMismatch (coerce requiredLength) (coerce inputLength)
```
Kleidukos, line 91: I'm not sure this function belongs to this module. It probably should be living in `sel/src/Sel/PublicKey/Internal/Signature.hs`. Also I do not the View Pattern buys us much here, it's probably overkill. :)

PR #182, opened 2025-02-17
Consistently rename ciphertext-related types and functions in sel

### LINKED ISSUE #181: Naming of `Sel.SecretKey.Cipher.Hash`
In `sel`, why do we refer to the ciphertext + authentication tag `secretbox` output as a "hash"? At a glance, this seems like it could be misleading.

Would you be open to a PR that renames this to `AuthenticatedCiphertext` or something more along those lines?

intricate:
Resolves #181

> In `sel`, why do we refer to the ciphertext + authentication tag `secretbox` output as a "hash"? At a glance, this seems like it could be misleading.

### DIFF
--- a/sel/src/Sel/SecretKey/Cipher.hs
+++ b/sel/src/Sel/SecretKey/Cipher.hs
@@ -34,12 +34,12 @@ module Sel.SecretKey.Cipher
   , nonceFromHexByteString
   , nonceToHexByteString
 
-    -- ** Hash
-  , Hash
-  , hashFromHexByteString
-  , hashToBinary
-  , hashToHexByteString
-  , hashToHexText
+    -- ** Ciphertext
+  , Ciphertext
+  , ciphertextFromHexByteString
+  , ciphertextToBinary
+  , ciphertextToHexByteString
+  , ciphertextToHexText
   ) where
 
 import Control.Monad (void, when)
@@ -75,7 +75,7 @@ import Sel.Internal.Sodium (binaryToHex)
 -- $introduction
 -- "Authenticated Encryption" uses a secret key along with a single-use number
 -- called a "nonce" to encrypt a message.
--- The resulting hash is accompanied by an authentication tag.
+-- The resulting ciphertext is accompanied by an authentication tag.
 --
 -- Encryption is done with the XSalsa20 stream cipher and authentication is done with the
 -- Poly1305 MAC hash.
@@ -213,7 +213,7 @@ instance Show Nonce where
 -- | Generate a new random nonce.
 -- Only use it once per exchanged message.
 --
--- Do not use this outside of hash creation!
+-- Do not use this outside of ciphertext creation!
 newNonce :: IO Nonce
 newNonce = do
   (fPtr :: ForeignPtr CUChar) <- Foreign.mallocForeignPtrBytes (fromIntegral cryptoSecretboxNonceBytes)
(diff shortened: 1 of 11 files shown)


sel/src/Sel/SecretKey/Cipher.hs
```diff
@@ -255,16 +255,16 @@ nonceToHexByteString (Nonce nonceForeignPtr) =
 -- | A ciphertext consisting of an encrypted message and an authentication tag.
 --
 -- @since 0.0.1.0
-data Hash = Hash
+data Ciphertext = Ciphertext
```
intricate, line 258: Do we prefer `Ciphertext` or `CipherText`?

I noticed a little bit of inconsistency:

In `Sel.PublicKey.Cipher` and `Sel.SecretKey.Stream`, we have two types named `CipherText`. However, in `Sel.SecretKey.Stream`, there are functions where this capitalization is not consistent: `ciphertextFromHexByteString`, `ciphertextToBinary`, etc.

Just let me know which way you guys prefer and I can make the changes throughout the project, if you like.

intricate: I've personally always used "ciphertext" as a single word but, of course, I'd like to comply with the API that you prefer.

sel/src/Sel/SecretKey/Cipher.hs
```diff
@@ -293,69 +293,70 @@ instance Ord Hash where
...
+instance Show Ciphertext where
+  show = BS.unpackChars . ciphertextToHexByteString
 
--- | Create a 'Hash' from a binary 'StrictByteString' that you have obtained on your own,
--- usually from the network or disk. It must be a valid hash built from the concatenation
--- of the encrypted message and the authentication tag.
+-- | Create a 'Ciphertext' from a hexadecimal-encoded 'StrictByteString' that
+-- you have obtained on your own, usually from the network or disk. It must be
+-- a valid ciphertext built from the concatenation of the encrypted message and
+-- the authentication tag.
```
intricate, line 308: Any idea what line width we're generally targeting? I changed this to 80 chars when updating the docs, but I'm happy to comply with whatever you prefer.

Kleidukos: Ideally less than 120 columns so that on a split screen, lines don't have to be wrapped. :)

sel/src/Sel/SecretKey/Cipher.hs
Kleidukos: Wikipedia says `Ciphertext`, let's stick with it!

Kleidukos:
Since this a breaking change of the API, could you introduce a changelog entry (https://github.com/haskell-cryptography/libsodium-bindings/blob/main/sel/CHANGELOG.md) for v1.0.0.0 ? And bump the version in `sel.cabal` as well?

sel/src/Sel/SecretKey/Cipher.hs
intricate: Sounds good. 

I'm thinking to follow up with another PR to change `CipherText` to `Ciphertext` in other modules. 

But, if you prefer it to be done in this PR, I'll just append the commits to this branch.

Kleidukos: I'd rather you did in this PR, at least there's one place to look at. :)

intricate: Done

intricate:
hi all, not to be a bother, but I just wanted to check in on this :grin:

Kleidukos:
@mergifyIO rebase

Kleidukos (approved):
@intricate Thank you for your patience! This is excellent work and I'm really honoured you felt that this project was worth contributing to. :bow:

intricate:
@Kleidukos: No problem at all. Happy to help out and continue contributing.

PR #183, opened 2025-02-19
Add support for additional data (AD) in Sel.SecretKey.Stream

intricate:
Add support for specifying AAD in `Sel.SecretKey.Stream`.

### DIFF
--- a/sel/src/Sel/SecretKey/Stream.hs
+++ b/sel/src/Sel/SecretKey/Stream.hs
@@ -12,7 +12,7 @@
 -- |
 --
 -- Module: Sel.SecretKey.Stream
--- Description: Encrypted Streams with ChaCha20Poly1305
+-- Description: Encrypted Streams with XChaCha20-Poly1305
 -- Copyright: (C) Hécate Moonlight 2024
 -- License: BSD-3-Clause
 -- Maintainer: The Haskell Cryptography Group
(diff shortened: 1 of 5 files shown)


sel/src/Sel/SecretKey/Stream.hs
```diff
@@ -461,6 +479,47 @@ messageTagToConstant = \case
...
+  deriving stock (Show, Eq)
+
+-- | Construct an 'AdditionalData' value from a hexadecimal-encoded
+-- 'StrictByteString' that you have obtained on your own, usually from the
+-- network or disk.
+additionalDataFromHexByteString
+  :: Base16 StrictByteString
+  -> Either AdditionalDataHexDecodingError AdditionalData
+additionalDataFromHexByteString hexBs =
+  case Base16.decodeBase16Untyped (Base16.extractBase16 hexBs) of
```
intricate, line 519: This pattern of calling `decodeBase16Untyped . extractBase16` instead of `decodeBase16` on a `Base16 a` value confuses me a bit (https://github.com/haskell-cryptography/libsodium-bindings/pull/144#discussion_r1962461202), but I've done this here to be consistent with similar functions in this module.

sel/src/Sel/SecretKey/Stream.hs
```diff
@@ -51,6 +51,13 @@ module Sel.SecretKey.Stream
     -- ** Message Tags
   , MessageTag (..)
 
+    -- ** Additional data (AD)
+  , AdditionalData (..)
```
intricate, line 55: Thoughts on me exposing this constructor? Or would you prefer a function like `additionalDataFromBinary` or something like that?

Kleidukos:
@intricate Thanks! I'll take time to review and think about your comment in #144 :)

sel/src/Sel/SecretKey/Stream.hs
Kleidukos: If the constructor is not needed to use the API and would just be used for access, then you can just expose `additionalDataFromBinary` indeed. :)

Kleidukos (changes requested):
You're missing a changelog entry and we're good!

Kleidukos (approved):
Thanks a lot!

PR #188, opened 2025-05-18
Fix `sel` HMAC tests for SHA-512 and SHA-512/256

intricate:
`Test.HMAC` was importing the wrong modules for SHA-512 and SHA-512/256.


Kleidukos (approved):

Kleidukos:
Thank you very much :pray:

PR #189, opened 2025-05-18
Define some cryptographic functions in `sel` as pure

### LINKED ISSUE #187: Pure cryptographic functions in `sel`
I wanted to get some opinions on this. 

Might it not be safe to define a few of the cryptographic functions in `sel` as pure (using `unsafeDupablePerformIO`)?

Some examples:

- `Sel.Hashing.hashByteString`
- `Sel.Hashing.SHA256.hashByteString`
- `Sel.Hashing.SHA512.hashByteString`
- `Sel.Hashing.Short.hashByteString`
- `Sel.Hashing.Password.hashByteStringWithParams`
- `Sel.HMAC.SHA256.authenticate`
- `Sel.HMAC.SHA512.authenticate`
- `Sel.HMAC.SHA512_256.authenticate`
- `Sel.SecretKey.Authentication.authenticate`
- `Sel.PublicKey.Signature.signMessage`

intricate:
Closes #187 

Defines the following functions as pure:

- `Sel.Hashing.hashByteString`
- `Sel.Hashing.SHA256.hashByteString`
- `Sel.Hashing.SHA512.hashByteString`
- `Sel.Hashing.Short.hashByteString`
- `Sel.Hashing.Password.hashByteStringWithParams`
- `Sel.HMAC.SHA256.authenticate`
- `Sel.HMAC.SHA512.authenticate`
- `Sel.HMAC.SHA512_256.authenticate`
- `Sel.SecretKey.Authentication.authenticate`

### DIFF
--- a/sel/src/Sel/PublicKey/Signature.hs
+++ b/sel/src/Sel/PublicKey/Signature.hs
@@ -162,6 +162,15 @@ generateKeyPair = do
 
 -- | Sign a message.
 --
+-- Note that, if @libsodium@ is compiled with the @ED25519_NONDETERMINISTIC@
+-- macro defined, this function will produce non-deterministic (but also
+-- non-standard) Ed25519 signatures. If @libsodium@ hasn't been compiled with
+-- the @ED25519_NONDETERMINISTIC@ macro defined, it's safe to call this
+-- function in a pure context with 'unsafeDupablePerformIO'.
+--
+-- For more information, see the
+-- [@libsodium@ docs](https://doc.libsodium.org/public-key_cryptography/public-key_signatures#notes).
+--
 -- @since 0.0.1.0
 signMessage :: StrictByteString -> SecretKey -> IO SignedMessage
 signMessage message (SecretKey skFPtr) =
--- a/sel/src/Sel/HMAC/SHA256.hs
+++ b/sel/src/Sel/HMAC/SHA256.hs
@@ -117,9 +117,9 @@ authenticate
   -- ^ Message to authenticate
   -> AuthenticationKey
   -- ^ Secret key for authentication
-  -> IO AuthenticationTag
+  -> AuthenticationTag
   -- ^ Cryptographic tag for authentication
-authenticate message (AuthenticationKey authenticationKeyForeignPtr) =
+authenticate message (AuthenticationKey authenticationKeyForeignPtr) = unsafeDupablePerformIO $
   BS.unsafeUseAsCStringLen message $ \(cString, cStringLen) -> do
     authenticationTagForeignPtr <-
       Foreign.mallocForeignPtrBytes
--- a/sel/src/Sel/HMAC/SHA512.hs
+++ b/sel/src/Sel/HMAC/SHA512.hs
@@ -117,9 +117,9 @@ authenticate
   -- ^ Message to authenticate
   -> AuthenticationKey
   -- ^ Secret key for authentication
-  -> IO AuthenticationTag
+  -> AuthenticationTag
   -- ^ Cryptographic tag for authentication
-authenticate message (AuthenticationKey authenticationKeyForeignPtr) =
+authenticate message (AuthenticationKey authenticationKeyForeignPtr) = unsafeDupablePerformIO $
   BS.unsafeUseAsCStringLen message $ \(cString, cStringLen) -> do
     authenticationTagForeignPtr <-
       Foreign.mallocForeignPtrBytes
--- a/sel/src/Sel/HMAC/SHA512_256.hs
+++ b/sel/src/Sel/HMAC/SHA512_256.hs
@@ -116,9 +116,9 @@ authenticate
   -- ^ Message to authenticate
   -> AuthenticationKey
   -- ^ Secret key for authentication
-  -> IO AuthenticationTag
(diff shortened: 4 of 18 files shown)


sel/src/Sel/PublicKey/Signature.hs
```diff
@@ -163,8 +163,8 @@ generateKeyPair = do
 -- | Sign a message.
 --
 -- @since 0.0.1.0
-signMessage :: StrictByteString -> SecretKey -> IO SignedMessage
-signMessage message (SecretKey skFPtr) =
+signMessage :: StrictByteString -> SecretKey -> SignedMessage
+signMessage message (SecretKey skFPtr) = unsafeDupablePerformIO $
```
intricate, line 167: This one is a bit questionable because of [this snippet in `libsodium`](https://github.com/jedisct1/libsodium/blob/3e12b517ee42eb4993d19611df325a01104709cc/src/libsodium/crypto_sign/ed25519/ref10/sign.c#L63-L67):

```c
#ifdef ED25519_NONDETERMINISTIC
    _crypto_sign_ed25519_synthetic_r_hv(&hs, nonce /* tmp */, az);
#else
    crypto_hash_sha512_update(&hs, az + 32, 32);
#endif
```

As mentioned in the[ `libsodium` docs](https://doc.libsodium.org/public-key_cryptography/public-key_signatures#notes), non-deterministic `Ed25519` signatures are non-standard, so it's doubtful that most people are building the library with this macro defined.

However, it's totally possible that it is defined. So, unfortunately, maybe it isn't safe to define this as pure :disappointed:

Kleidukos: Hmm yeah, maybe let's keep that one with `IO`, and we'll indicate that in the absence of `ED25519_NONDETERMINISTIC` it's ± safe to use `unsafeDupablePerformIO`

Kleidukos (approved):

Kleidukos:
@intricate I'll merge when CI is green, good job!

intricate:
Outside of `Sel.PublicKey.Signature.signMessage`, it seems safe to define all of these functions as pure. 

As we know, `libsodium` makes heavy use of "out" parameters, so technically none of this stuff is _truly_ pure. However, apart from mutating buffers and whatnot (of course, **not** mutating any of the _"in"_ parameters), there doesn't appear to be anything else that's secretly stateful. So I think it's probably safe to just call these functions pure.

One thing could be an additional concern in defining these as pure, is the fact that, in data types such as `AuthenticationTag`, `Hash`, etc., we hold on to `ForeignPtr`s as opposed to converting values back to `ByteString`s in Haskell-land. 

This currently seems okay since none of these `ForeignPtr` values are exposed from their modules, but maybe it could be seen as "bad" to say these functions are pure when _technically_ it's possible to dereference and mutate the output values via these pointers.

intricate:
Oops, also forgot to update the package API golden files.

intricate:
> One thing could be an additional concern in defining these as pure, is the fact that, in data types such as `AuthenticationTag`, `Hash`, etc., we hold on to `ForeignPtr`s as opposed to converting values back to `ByteString`s in Haskell-land. 

eh, nvm. I think I'm reading into this too much. Technically, someone could also just call [`Data.ByteString.Internal.toForeignPtr`](https://hackage-content.haskell.org/package/bytestring-0.12.2.0/docs/Data-ByteString-Internal.html#v:toForeignPtr) on a `ByteString` and get that `ForeignPtr` too.

So nevermind my comment lol.

sel/src/Sel/PublicKey/Signature.hs
```diff
@@ -162,6 +162,15 @@ generateKeyPair = do
...
 --
+-- Note that, if @libsodium@ is compiled with the @ED25519_NONDETERMINISTIC@
+-- macro defined, this function will produce non-deterministic (but also
+-- non-standard) Ed25519 signatures. If @libsodium@ hasn't been compiled with
+-- the @ED25519_NONDETERMINISTIC@ macro defined, it's safe to call this
+-- function in a pure context with 'unsafeDupablePerformIO'.
+--
+-- For more information, see the
+-- [@libsodium@ docs](https://doc.libsodium.org/public-key_cryptography/public-key_signatures#notes).
+--
```
intricate, line 173: @Kleidukos: how do you feel about this comment?

intricate:
_Deleted my previous comment to avoid muddying the waters with a non-issue that I brought up :grin:_

---

Outside of `Sel.PublicKey.Signature.signMessage`, it seems safe to define all of these functions as pure. 

As we know, `libsodium` makes heavy use of "out" parameters, so technically none of this stuff is _truly_ pure. However, apart from mutating buffers and whatnot (of course, **not** mutating any of the _"in"_ parameters), there doesn't appear to be anything else that's secretly stateful. So I think it's probably safe to just call these functions pure.

sel/src/Sel/PublicKey/Signature.hs
Kleidukos: Excellent.

Kleidukos (approved):

Kleidukos:
@intricate Thank you very much for you recent contributions. I believe this was the last PR needed to release a new major version of `sel`. :)

